Issue #8, opened 2019-02-07
ROPE: default


DominiqueMakowski:
Enable the ROPE parameter to "default", which will be -0.1 0.1 for rope.numeric and 0.1 * SD(response) for models... But what in the case of logistic models? What's the best way to return the ROPE bounds?

DominiqueMakowski:
Option: Return a list containing the rope percentage, the CI and the bounds, with a wrapping print method

PR #9, opened 2019-02-07
add support for models

DominiqueMakowski:
- [x] Support
- [x] Testing
- [ ] Rope default #8


.Rproj.user/B5F94F77/sources/s-C9E6AA8B/83198F1F-contents
```diff
@@ -29,19 +29,45 @@
...
 #' @export
-rope <- function(posterior, bounds = c(-0.1, 0.1), CI = 90, verbose = TRUE) {
+rope <- function(posterior, bounds = "default", CI = 90, verbose = TRUE) {
   UseMethod("rope")
 }
 
 
+#' @method as.numeric rope
+#' @export
+as.numeric.rope <- function(x, ...){
```
DominiqueMakowski: this doens't work 😭

DominiqueMakowski:
@strengejacke I've implemented a first iteration of model supports, as well as ROPE that adapts to the response's SD if a model is provided (hence, the package now imports insight). However, I do have a problem with a the `as.numeric.rope` method which doesn't work, so for now I've added the `as_numeric_rope()` function to temporarily mimic the previous one.... 

Do you have any idea on why it doens't work? 

And what do you think about the current implementation of model's support

strengejacke:
I would add `.RHistory` and `.RProj*` to the gitignore-files, so these files don't appear in the commits. This makes it easier to review PRs. See e.g. here https://github.com/easystats/insight/blob/master/.gitignore

.Rproj.user/B5F94F77/sources/s-C9E6AA8B/83198F1F-contents
strengejacke: I just downloaded and played around with the dev-branch, and I am confused as well. What I've read from `?as.numeric`:

> as.numeric is a generic function, but S3 methods must be written for as.double. It is identical to as.double.

So I tried an `as.double.rope`,  but had no success, either...

strengejacke:
> And what do you think about the current implementation of model's support

The implementation looks good! I would probably revise the `print()`-method for list-return-values.

DominiqueMakowski:
> I would probably revise the `print()`-method for list-return-values.

How would this look?

strengejacke:
E.g. for `equivalence_test()`, the current output is:

```r
> equivalence_test(model)
$`(Intercept)`
[1] "rejected"

$wt
[1] "rejected"

$cyl
[1] "rejected"

$sigma
[1] "rejected"
```

For models, I would a) give information on the effect size, b) mention the ROPE values and c) maybe a "tabular" print, one liner per parameter (similar to what `equi_test()` does). Unless you want other packages to be responsible for proper printing, but these are already "real tests", so I would find a "human readable" way for the printout.

.Rproj.user/B5F94F77/sources/s-C9E6AA8B/83198F1F-contents
strengejacke: Ok, should work now, with `as.double.rope()` - in the last try, I forgot to update the NAMESPACE. ^^

strengejacke:
I would say you can merge this PR now... The core functionality is in, and further refinement, print methods etc can be done in a next step.

PR #10, opened 2019-02-08
outputs as dataframes

DominiqueMakowski:
This update changes the behaviour of almost all functions in order to return dataframes instead of lists. I guess this makes sense from a user perspective. 

I do expect that bayestestR to be used mainly by other packages to compute indices for further analyses (or visualisation). At the same time, people that want to actually *use* these functions for their results will likely use higher levels packages (parameters, report etc.) that incorporate good defaults and diegetically present and wrap up more information. 

Thus, I think it makes sense to implement pretty formatting in report or such rather than at this level, that would mostly add noise and unexpected behaviour for those wanting to exploit the functions within a longer pipeline. Here would be instead of the place to clearly and thoroughly explain the different indices and metrics, their use and so on, but return the information in a neutral (in regards to formatting) manner.

However, there are still a few printing methods implemented for some functions, I don't know if we should keep them or remove them (i.e., move them to other packages) to be consistent with this perspective.

@strengejacke what do you think of this perspective?


R/equivalence_test.R
```diff
@@ -36,30 +38,51 @@ equivalence_test <- function(posterior, bounds = "default", CI = 90, verbose = T
...
+  return(names(coef(model)))
 }
 
+#' TEMPORARY
+#' @importFrom stats coef
+#' @param model model
 #' @export
-equivalence_test.brmsfit <- function(posterior, bounds = "default", CI = 90, verbose = TRUE) {
-  sapply(as.data.frame(posterior), equivalence_test, bounds = bounds, CI = CI, verbose = verbose, simplify = FALSE)
+get_parameters <- function(model) {
```
DominiqueMakowski: get_parameters and find_parameters were temporarily put here (related to [this](https://github.com/easystats/insight/issues/24))

strengejacke: Have you updated the gitignore file on your computer? I'll take a look at this later, it's indeed a bit too messy to inspect from mobile phone. ;-)

strengejacke:
> However, there are still a few printing methods implemented for some functions, I don't know if we should keep them or remove them (i.e., move them to other packages) to be consistent with this perspective.

I would consider this package at a mid level, which can also directly be used by end users. It cobbles computes tests, so I would expect a human readable output from the results. 

Else we would end up with another package having the same function that only provides a print-method. What do you think?

DominiqueMakowski:
That's true. Although the dataframe outputs, being more flexible (better dealing with vectors of parameters), makes it less straightforward to implement clever printing, we could give it a try with some neutral formatting! Feel free to add some ideas or examples ☺️

strengejacke:
"cobbles" was inserted by auto-completion in my above comment, I removed it... ^^

R/utils_flatten_list.R
```diff
@@ -0,0 +1,31 @@
...
+flatten_list <- function(list, name = "name") {
+  if (length(list) == 1) {
+    return(list[[1]])
+  } else
+  if (all(sapply(list, is.data.frame))) {
+    if (is.null(names(list))) {
+      return(as.data.frame(t(sapply(list, rbind))))
+    } else {
+      tryCatch({
+        df <- data.frame()
```
strengejacke: Don't use function names as variable names ;-)

strengejacke:
Two things come into my mind:

1) We should also think about naming conventions here. `CI` is an upper case argument, which could be renamed into something lower case ( `ci`, `ci.level`, ...)
2) Should we break the convention of providing arguments like `CI` as numeric value between 0 and 1? Currently, you have to pass a value from 0 to 100. I think most users will intuitively write something like `.95` for a 95% interval (this is not only the convention, but also 95% = 0.95). I think this causes a lot of confusion.

strengejacke:
In general I think having data frames is more convenient. As the returned elements all have the same lengths, they can be put into a data frame, and having a data frame makes it easier to extract either the data for a parameter (a row) or for a statistic (a column).

Returning the data as it currently is implemented (data frames) is the way to go, I would say. I can start working on a `print()`-method in another branch. Do we want colored output like if we use the _crayon_ package? We can easily re-implement such functions to keep dependency at a minimum.

strengejacke:
I would say the _print()_ method should be located in bayestestR, while specific demands on plotting / text / tabular output should go into report.

DominiqueMakowski:
For the CI argument, I guess it's ok to rename it `ci` (I am just sooo used to `CI` but hey, it's good to come out of my comfort zone 😄). However, I would suggest keepking the CI as a percentage, as in my many functions (e.g., `confint`), the related parameter is named `prob` and corresponds to the probability (between 0 and 1). However, the CI parameter refers to the credible interval, more often expressed in percentage (e.g., 95% CI), thus I think it would be ok to leave it this way... And it's always one or two key strokes less to type... (`0.` 👍)

DominiqueMakowski:
> Do we want colored output like if we use the _crayon_ package? We can easily re-implement such functions to keep dependency at a minimum.

I've never worked with crayon, so I don't know if it's complicated, but yes it could be nice!

strengejacke:
Try `cat("\033[31measystats\033[39m")`. If your console or RStudio supports colored console-output, you'll see a red-colored _easystats_. We just need a function that checks wether the current user-setup supports colored output (we can steal this from crayon), and if so, wrap the ansi-code for the color around the string, else just print the "pure" string w/o color-code.

strengejacke:
> However, the CI parameter refers to the credible interval, more often expressed in percentage (e.g., 95% CI)

True, but you don't type the %-sign - so, 95 as _percent_ is equal to 0.95. See `?stats::confint`: The `level`-argument is between 0 and 1, and the help-file says: "These will be labelled as (1-level)/2 and 1 - (1-level)/2 in % (by default 2.5% and 97.5%)." So the value of `level`, between 0 and 1, referes to a percentage value.

I don't mind using an own convention, but we should be careful when it might become different from typical user habits. I personally have no experiences with using values from 0 to 100 in my packages, so I'm not sure how users will perceive this determination, or if it's a problem at all or even better.

DominiqueMakowski:
Ok, you've convinced me 😄

strengejacke:
btw, the vignettes in this package are great! And very helpful in understanding the purpose of this package. But coming back to the topic: Can we merge this PR or is there something you're missing?

DominiqueMakowski:
(Render unto Caesar - the vignettes were **strongly** inspired by sjstats' 😏)

There is the `find_*` / `get_parameters` placeholders that I'd like to get rid off in order to re-enable brms testing and support... do you have any quick/temporary idea for it?

strengejacke:
> do you have any quick/temporary idea for it?

Yes, add a `## TODO remove later` comment, and once we have a better solution, remove the functions ;-)
Do you know the TODOr package? https://github.com/dokato/todor That scans your package for `## TODO ...`, `## FIXME ...` etc. Quite useful for such things. We should probably move the get/find parameter into insight for now, and maybe later move them into a parameter-or-whatever package, if necessary.

DominiqueMakowski:
Thanks for the link to `todor`! 

I think we can merge, and once parameters insights will work for brms, we will uncomment examples and tests here. thanks!!

DominiqueMakowski:
Travis fails but I believe it related to the get_parameters that are not in master for insight, so will merge anyway

PR #14, opened 2019-02-10
authors

DominiqueMakowski:
I've taken the initiative of adding you in the citation info and temporary reference (which hopefully will be updated). However, I would understand that this might be a little early and that you don't want your name to be associated with the package's content (I am thinking about the vignettes) in this state 😅 . Let me know!


strengejacke:
totally fine!

PR #23, opened 2019-03-04
revise rope-bounds for binomial models

strengejacke:


strengejacke:
Are you ok with these changes? It comes close to what Kruschke suggests in the appendix of his paper (Kruschke JK. Rejecting or Accepting Parameter Values in Bayesian Estimation. Advances in Methods and Practices in Psychological Science. 2018; doi: 10.1177/2515245918771304 ), and is probably a better approximation of the rope-bounds compared to the fixed value of `(pi / sqrt(3))`. According to Kruschke, the rope-bounds for binomial models depends on the proportion of "successes", which is reflected in this code-revision.

DominiqueMakowski:
I am ok! We should add this in the docstrings.

DominiqueMakowski:
(I would be personally interested in a comparison between the logistic linear-cohen's-d-direct-equivalent and the general binomial methods, to see if it changes much)

strengejacke:
``` r
prob_resp <- seq(.05, .5, .05)
eff_size <- prob_resp / pi
effect_size_d <- (stats::qlogis(prob_resp + eff_size) - stats::qlogis(prob_resp - eff_size)) / 4
cbind(proportion_success = prob_resp, log_odds_equivalent_to_01_d = effect_size_d)
#>       proportion_success log_odds_equivalent_to_01_d
#>  [1,]               0.05                   0.1732600
#>  [2,]               0.10                   0.1825739
#>  [3,]               0.15                   0.1929984
#>  [4,]               0.20                   0.2047557
#>  [5,]               0.25                   0.2181347
#>  [6,]               0.30                   0.2335198
#>  [7,]               0.35                   0.2514358
#>  [8,]               0.40                   0.2726231
#>  [9,]               0.45                   0.2981704
#> [10,]               0.50                   0.3297653

0.1 * (pi / sqrt(3))
#> [1] 0.1813799
```

<sup>Created on 2019-03-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

As you can see, the "fixed" effect size `0.1 * (pi / sqrt(3))` ~ .18 and corresponds to a proportion of ~ 10% of successes in the outcome. I think most realistic scenarios are a proportion of "success" of about 5% to 25% - so the new proposal is not far from the "fixed" solution.

DominiqueMakowski:
interesting

strengejacke:
If I read the relevant paragraph again:

> Formally, denote the predicted probability of "success" as p and the log-odds as `logit(p) = log(p(/1-p))`. We start by dening a negligible change in p. Toward that end, suppose we are conducting a political poll in a population with preferences for candidates A and B split approximately 50/50, and we are
predicting preference as a function of income. Suppose we specify that a negligible change
in p over the main range of x is p = 0:50 +/- 0:03. If the main range of x is Mx +/- 2Sx
(as in the example of linear regression in the main article), then a negligible standardized
regression coefficient has a ROPE of `bx < (logit(0:53) - logit(0:47))=4 ~ 0:06`. Suppose
instead we are studying occurrence of heart attack during a ve-year window as predicted
by the patient's average systolic blood pressure, and we specify that a negligible change in
p over the main range of p = 0:020 +/- 0:002. Then a negligible standardized regression
coeffiient has a ROPE of `bj < (logit(0:022) - logit(0:018))=4 ~ 0:05`.

This makes me doubt if my numbers aren't to high, and we should actually divide by `pi^2` instead of just pi:

``` r
prob_resp <- seq(.05, .5, .05)
eff_size <- prob_resp / pi^2
effect_size_d <- (stats::qlogis(prob_resp + eff_size) - stats::qlogis(prob_resp - eff_size)) / 4
cbind(proportion_success = prob_resp, log_odds_equivalent_to_01_d = effect_size_d)
#>       proportion_success log_odds_equivalent_to_01_d
#>  [1,]               0.05                  0.05350140
#>  [2,]               0.10                  0.05646422
#>  [3,]               0.15                  0.05977609
#>  [4,]               0.20                  0.06350289
#>  [5,]               0.25                  0.06772832
#>  [6,]               0.30                  0.07256037
#>  [7,]               0.35                  0.07814092
#>  [8,]               0.40                  0.08466026
#>  [9,]               0.45                  0.09237995
#> [10,]               0.50                  0.10167006
```

<sup>Created on 2019-03-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This is still more conservative than Kruschke's suggestion, but I think this is better than "p-hacking".

strengejacke:
I think I revert back the changes to `/ pi` instead of `/ pi^2`. Looking at my "real-life-models", I don't expect an odds ratio of 1.03 as "undecided", but rather as practical equivalent. With `pi`, it's considered as practical equivalent, while `pi^2` denotes this odds ratio as "undecided".

DominiqueMakowski:
I think that these are two separate questions, a) finding a consistent default value across models and b) discussing whether this value is conservative enough, and making possible adjustments to be made equivalently for all models.

It would be great to have comparable linear with binomial models and run some simulation to estimate a consistent threshold across models to first answer a)

DominiqueMakowski:
> I think I revert back the changes to / pi instead of / pi^2.

Ok!

Issue #28, opened 2019-03-06
Function for credible intervals


strengejacke:
I think we should also add a function that returns the credible intervals (not HDI) as well, for convenience reasons, where the output meets our needs.

In case of transformation, eg having odds ratios, you can also exp the CI, but not the HDI.

See https://discourse.mc-stan.org/t/exponentiation-or-transformation-of-point-estimates/7848

DominiqueMakowski:
That's a very good point indeed!!! (it struck me as obvious upon reading your issue and I now I wonder how is it possible that I had never thought of it 😅)

DominiqueMakowski:
If I understand correctly, CIs based on percentile (equal-tailed intervals) would be more suited for binomial data. However, CIs based on the HDI could still be better for other types of models... is that right?

We could have a function, e.g. `ci()` with a parameter `method="hdi"` or `method="percentile"` that would return the two versions of what can be further referred to as credible interval. Also, a "default" argument would compute the most appropriate CIs based on the data/the model at hand. Would that do the trick?

strengejacke:
Having one function with `method`-argument would maybe indeed be a good solution.

> However, CIs based on the HDI could still be better for other types of models... is that right?

You can use either CI or HDI for any model, however, HDI is useful / appropriate if it's reported on the scale of the linear predictor. For instance, for a logistic regression, reporting log-odds and HDI is fine. Reporting odds ratios with exponentiated HDI may be misleading, because in one case (log-odds scale) the HDI does _not_ include "zero", while on the odds ratio scale, it may include "zero" (resp. "one" in this case). In frequentist terms: on the log-scale, the point-estimate is "significant", while it's no longer significant on the exponentiated scale.

So for functions like `equivalence_test()`, we should keep using HDI. But if you report estimates on the exponentiated scale, the credible interval (CI) is more appropiate, because both on log- and exp-scale, the CI "preserves" the "(non-)significance".

strengejacke:
Closing in favour of #30

PR #29, opened 2019-03-06
updates

strengejacke:


strengejacke:
@DominiqueMakowski The 2nd commit (reorganize hdi()) is a rather large revision of the `hdi()`-function. I first wanted to work on https://github.com/easystats/bayestestR/issues/28, however, I think `hdi()` needs some revisions. The function currently only returns the HDI for the fixed effects of the conditional component only. For more complex models, especially from brms, a large proportion of information on other components and random effects is missing, which might be interesting, though.

I revised the function, so it now returns the HDI for fixed/random and conditional/zeroinflated components, and does some cleaning for the parameter names. As a side-effect, we have now two more columns in the returned data frame, `Group` and `Component`, which indicate to which model-part a parameter belongs to.

Would you say this is a good approach, or do you feel this is kind of overload of information?

strengejacke:
Once HDI is in the shape we want it to have, I will start working on the `ci()`-function, which either returns percentiles or HDI.

strengejacke:
We maybe should think of a similar approach for `rope()` as well...

strengejacke:
The next commits fixes `rope_bounds()`, so it works with multivariate response models. A bit more difficult is to fix `rope()` and `equivalence_test()` as well...

DominiqueMakowski:
> I revised the function, so it now returns the HDI for fixed/random and conditional/zeroinflated components, and does some cleaning for the parameter names. As a side-effect, we have now two more columns in the returned data frame, Group and Component, which indicate to which model-part a parameter belongs to. 

Hum that's another design issue. Maybe we could use an insight-like syntax, with an argument `effects` and `component`, defaulting to "fixed" and "all", that would make a selection in the returned data. If `effects` and `components` is not `all` or list, then remove the appropriate column (e.g., if the user asked for `fixed` only, do not return the `Group` column. In this way, we could increase the consistentcy between insight and bayestestR might be a good thing.

What do you think?

strengejacke:
Sounds good! Do we want a list for each component (so we don't need extra column), or an integrated data frame (as we have it now in this PR)?

DominiqueMakowski:
I would say that an integrated data.frame is ok

strengejacke:
The commit https://github.com/easystats/bayestestR/pull/29/commits/612dacee3911b3931f2bf90a939adc23fa520f5a 

- revises the print()-methode for `rope()` when a model is passed as argument.
- updates the reference for `equivalence_test()`
- defaults the ci-level for `equivalence_test()` to 95%, as this is the suggested value from Kruschke (no matter if 90% HDI is more stable, but this is what he suggested, and this is being more conservative, leading to less parameters outside the ROPE, which is actually good in terms of "p-hacking")

strengejacke:
- [ ] I'll check the new `hdi()` for some models tonight, and if it works, we may extend the arguments `effects` and `component` to `rope()` as well.
- [ ] The print()-method for `equivalence_test()` for models should be revised, so it does not print redundant information, but includes useful information (HDI for each parameter) instead.
- [ ] Then, the `ci()`-method can be targeted, and this PR is "complete".

Would you say that we should have both `ci()` and `hdi()`, or as you suggested, just `ci()` with `method`-argument? I like both variants. Having a separate `hdi()` method might be easier for users to find if they explicetly search for such a function...

DominiqueMakowski:
- I agree with having both ci and hdi.
- Ok for the default revising to 95% for now, but now that we have cleaner code and flexible code, we will be able to think about the simulation study to reevaluate and provide objective data about the comparison between the indices variants and parameters.
- Great work! I like how due to the jet lag, easystats is being worked on 24/24 ☺️

strengejacke:
> but now that we have cleaner code

I'm not sure if it's cleaner ;-) But once everything works, we can see if we need to restructure parts of the code to be more flexible and easier to maintain.

R/rope.R
```diff
@@ -113,14 +114,20 @@ rope.numeric <- function(posterior, bounds = "default", ci = .90, verbose = TRUE
...
+  } else if (!all(is.numeric(range)) | length(range) != 2) {
+    stop("`range` should be 'default' or a vector of 2 numeric values (e.g., c(-0.1, 0.1)).")
   }
-  list <- sapply(insight::get_parameters(posterior), rope, bounds = bounds, ci = ci, verbose = verbose, simplify = FALSE)
-  flatten_list(list, name = "Parameter")
+
+  list <- sapply(insight::get_parameters(posterior), rope, range = range, ci = ci, verbose = verbose, simplify = FALSE)
+  hdi_area <- do.call(rbind, lapply(list, attr, "HDI_area"))
+
+  r <- flatten_list(list, name = "Parameter")
```
strengejacke: @DominiqueMakowski This is not very nice, but `flatten_list()` drops attrbiutes, and I need to copy along the information from the HDI for `equivalence_test()`, to get a more informative output for `print.equivalence_test()`.

DominiqueMakowski: I understand, in the future, we might need to think about a robust and neat solving for this flatten_list thing, as it is also present in `parameters`.

R/equivalence_test.R
```diff
@@ -29,14 +29,19 @@
...
 }
 
 
 #' @export
-equivalence_test.numeric <- function(posterior, bounds = "default", ci = .95, verbose = TRUE) {
-  out <- as.data.frame(rope(posterior, bounds = bounds, ci = ci))
+equivalence_test.numeric <- function(posterior, range = "default", ci = .95, verbose = TRUE) {
+  if (length(ci) > 1) {
+    stop("'ci' needs be a single value.", call. = FALSE)
+  }
```
strengejacke: @DominiqueMakowski Line 46 gives a warning when CI is longer than 1 e.g. `ci = c(.6, .8)`. This will not pass through CRAN checks, that's why I wanted to make users aware that only length-one is allowed for `ci`. However, I just saw in the examples and tests, that for numeric vectors, `equivalence_test()` was intended by you to work with multiple `ci`-values. How to deal with this?

R/rope.R
```diff
@@ -74,7 +94,7 @@ rope.numeric <- function(posterior, range = "default", ci = .90, verbose = TRUE)
     .rope(posterior, range = range, ci = i, verbose = verbose)
   })
 
-  out <- flatten_list(rope_values)
+  out <- do.call(rbind, rope_values)
```
strengejacke: @DominiqueMakowski `flatten_list()` seems to return a data frame with _list_ columns in this case. I don't think this is the expected behaviour, so I changed this code.

R/equivalence_test.R
strengejacke: Ok, this one is actually resolved by this commit: https://github.com/easystats/bayestestR/pull/29/commits/1eaac43fa2e9f857dae4dcd9065e2e0a7f98dee0

I need to revise the `print()`-method to print out the necessary information for equivalence-tests with multiple CI.

strengejacke:
btw, feel free to comment on the style (visual appearance / output in the console) of the revised print()-methods. Propose any changes if you prefer a different "look".

R/rope.R
```diff
@@ -43,13 +43,20 @@ as.double.rope <- function(x, ...) {
...
 print.rope <- function(x, ...) {
-  cat(sprintf(
-    "%.2f%% of the %s%% CI is in ROPE [%.2f, %.2f]",
-    x$ROPE_Percentage,
-    x$CI,
-    x$ROPE_low,
-    x$ROPE_high
-  ))
+  if (is.data.frame(x)) {
+    cat(sprintf("# Proportions of samples inside the ROPE [%.2f, %.2f]\n\n", x$ROPE_low[1], x$ROPE_high[1]))
```
DominiqueMakowski: I think it might be worth to specify to what "samples" we refer to, i.e., "samples from the credible interval" or "samples from the posterior distribution" (in the case ci = 1). And indicate the percentage of CI even when there is no multiple CIs

DominiqueMakowski: Also, might be fancy to say "Proportion" in the case `length(ci) == 1` and "Proportions" otherwise

DominiqueMakowski: And add "**:**" at the end of this header?

R/equivalence_test.R
```diff
@@ -48,21 +53,42 @@ equivalence_test.numeric <- function(posterior, bounds = "default", ci = .95, ve
...
+  cat(sprintf("  ROPE: [%.2f %.2f]\n\n", x$ROPE_low[1], x$ROPE_high[1]))
+  ci <- x$CI[1]
+
+  x$ROPE_Percentage <- sprintf("%.2f%%", x$ROPE_Percentage)
+  x$HDI <- sprintf("[%.2f %.2f]", x$HDI_low, x$HDI_high)
+
+  x <- x[, intersect(c("Parameter", "ROPE_Equivalence", "ROPE_Percentage", "HDI"), colnames(x))]
+
+  colnames(x)[which(colnames(x) == "ROPE_Equivalence")] <- "H0"
+  colnames(x)[which(colnames(x) == "ROPE_Percentage")] <- "Inside ROPE"
```
DominiqueMakowski: It might be clearer adding "Proportion" or "Percentage" or "%", e.g; "% inside ROPE"

R/equivalence_test.R
```diff
@@ -48,21 +53,42 @@ equivalence_test.numeric <- function(posterior, bounds = "default", ci = .95, ve
...
+  colnames(x)[ncol(x)] <- sprintf("%i%% HDI", ci)
```
DominiqueMakowski: This doesn't work in the case of multiple CIs:
```r
equivalence_test(posterior = rnorm(1000, 1, 1), ci = c(.50, .99))
```
```
# Test for Practical Equivalence

  ROPE: [-0.10 0.10]

        H0 Inside ROPE     50% HDI
  rejected       0.00% [0.25 1.58]
 undecided       4.24% [0.25 1.58]
```
One workaround would be moving the desired CI proportion in a separate column, this way they could differ from row to row.

strengejacke:
Your suggestions sound great, I'll work on it tomorrow. According multiple CI in print out: see #30 it's on the to-do list. I thought about multiple "blocks" in the output, one per CI.

PR #44, opened 2019-03-15
CI and Documentation

DominiqueMakowski:


DominiqueMakowski:
I found it easier in the "get started" to put the report's output, just to showcase the ease of Bayesian usage. Then, in the next examples, we will go more in depth into the different underlying bayestestR's functions 😏

R/ci.R
```diff
@@ -0,0 +1,46 @@
...
+ci <- function(model, ci = .90, ...) {
+  UseMethod("ci")
+}
+
+
+
+#' @rdname ci
+#' @export
+ci.numeric <- function(model, ci = .90, verbose = TRUE, ...) {
+  hdi(model, ci=ci, verbose=verbose, ...)
```
strengejacke: I guess this is a placeholder? I think we can actually use something like `lapply(ci, stats::quantile, ...)` here, or not?

DominiqueMakowski: Yes, it was just to see how can I expand on this method in parameters without redefining it

As for the stuff in itself, I was thinking about a `method` argument, with "hdi" or "quantile" as possible values?

strengejacke: We already thought about the `method`-argument and a `ci()`-function. I like this implementation, so in the end, you can always end up calling `ci()`, even if you want `hdi()`, which gives more flexibility (still, I would keep the `hdi()` function because it's better to find then).

DominiqueMakowski: I agree!

R/print.hdi.R
```diff
@@ -0,0 +1,104 @@
...
+      cat("\n")
+    }
+  }
+}
+
+
+
+
+
+
```
strengejacke: I see you like empty lines... ;-)

DominiqueMakowski: haha... I guess so, it helps me see the different functions (especially when navigating through internals). An alternative could be adding  `#-----------------` between functions?

DominiqueMakowski:
@strengejacke do you want to try implementing the first iteration of `ci()`? I feel like your vision for this is more clear

strengejacke:
Yeah, can do that...

R/hdi.R
```diff
@@ -246,32 +246,13 @@ hdi.brmsfit <- function(posterior, ci = .90, effects = c("fixed", "random", "all
...
-
-  # find minimum of width differences, check for multiple minima
-  min_i <- which(window_width_diff == min(window_width_diff))
-  n_candies <- length(min_i)
-
-  if (n_candies > 1) {
-    if (any(diff(sort(min_i)) != 1)) {
-      if (verbose) {
-        warning("HDI: Identical densities found along different segments of the distribution, choosing rightmost.")
-      }
```
strengejacke: I _think_ this is an edge case that happens once in ten thousand years...

DominiqueMakowski: better having it in "just in case"?

PR #50, opened 2019-03-29
CRAN Release

DominiqueMakowski:
prepare for CRAN.


strengejacke:
Ok, insight update is on CRAN:
https://cran.r-project.org/web/packages/insight/index.html

It take some days until all binaries are compiled, then we should think about submitting bayestestR to CRAN.

1) Do you want to take over responsibility for this?
2) What would you say about the time frame, when will we be ready for submission?

DominiqueMakowski:
I can submit tomorrow or on Sunday, as soon as the tests pass (related to insight)

I still think that we should move the plotting dependencies and functions out, but we can keep them for v1 and them relocate them to the visualisation package (and of course reexport the methods here so the transition is seamless)

strengejacke:
> I still think that we should move the plotting dependencies and functions out

Yes, I agree! Yet, for now, I would like to have them available right now, because from a user-perspective, there's more benefit than harm with how it's currently implemented. It's primarily for our maintainance-workload to move the plot-functions into a separate package.

So if you have no big hesitance against it, I'd be happy to have in in bayestestR. :-) Also, since I'm working on 1-2 papers that use the equivalence_test(), so I can already cite our package. ^^

strengejacke:
> I can submit tomorrow or on Sunday, as soon as the tests pass (related to insight)

I think it'll take until mid of next week until the insight-update is released for all test-environments on CRAN, but we'll see.

strengejacke:
https://github.com/easystats/bayestestR/projects/1

DominiqueMakowski:
> I think it'll take until mid of next week until the insight-update is released for all test-environments on CRAN, but we'll see.

Ok, I'll upload as soon as possible then :)

strengejacke:
Can we merge this, so we have a "clean" repo, where we also can build the pkgdown-website? This would help me looking through the docs etc. if I can work in the master-branch.

I'm more one of these faction:
https://twitter.com/jimhester_/status/1110945976708145152?s=20

DominiqueMakowski:
merged :) I wanted the tests to pass, but that's minor since master was not passing either ^^

strengejacke:
Yes, tests for insight still don't pass, but that's an osx-issue, because the lfe-package is still not updated for osx (https://cran.r-project.org/web/checks/check_results_lfe.html). I'm a bit undecided about testing on OSX. on the one hand, tests fail when packages are not (yet) released for osx. on the other hand, it's a chance to at least test on osx, because, who is using osx anyway? ;-)

DominiqueMakowski:
Although we're not doing any black magic that would depend on the hardware architecture, so I don't understand how tests can pass one some platforms and not on others... Discrepancies happened to me also with my other package, and I was quite confused.

As for the philosophy of branches, I do see it as layers of stability, CRAN is the super stable in regard to its dependencies and testing, master would be quite stable but with a faster cycle release (i.e., for instance on CRAN there are like 0.4.2, 0.5.0, 0.6.1, 0.6.2, 0.7.0 while the master would see all of the intermediate ones). And the dev branch is the WIP branch with features in development. And I can conceive even further "feature" branches, that would be focused at one thing, then merged into dev, before being added to master... but that's for another time ^^

To be honest it was a real pain at the beginning for me switching between and understanding branches (and resolving random conflicts 😠) but somehow I feel like I am getting used to it, and it's almost agreeable to see us adding commits to the dev branch's PR... Anyway, I'll upload bayestestR to CRAN once the osx sources are there :)

PR #67, opened 2019-04-08
0.1.1

DominiqueMakowski:
- [x] `map_estimate` change


R/p_map.R
```diff
@@ -41,7 +41,7 @@ print.p_map <- function(x, ...) {
 #' @export
 p_map.numeric <- function(posterior, precision = 2^10, ...) {
   # Highest density point
-  map <- map_estimate(posterior, precision = precision, density = TRUE)$MAP_density
+  map <- attributes(map_estimate(posterior, precision = precision))$MAP_density
```
strengejacke: is `attributes` the correct function here? Shouldn't it rather be `map <- attr("MAP_density", map_estimate(posterior, precision = precision))$MAP_density)` or so?

strengejacke: No, it should be:
`attr(map, "MAP_density") <- map_estimate(posterior, precision = precision)$MAP_density`

DominiqueMakowski: No no, this is in `p_map` where I need to retrieve the density. The problematic part is in `map_estimate.R`

R/map_estimate.R
```diff
@@ -46,62 +39,67 @@ map_estimate <- function(x, ...) {
...
 
   hdp_x <- d$x[which.max(d$y)]
   hdp_y <- max(d$y)
 
-  out <- data.frame(
-    "MAP" = hdp_x,
-    "MAP_density" = hdp_y
-  )
+  out <- hdp_x
+  attr(out, "MAP_density") <- hdp_y
```
DominiqueMakowski: here

strengejacke: ok, I see.
I fixed the test now. For vectors, or in general objects with attributes, simply wrap the function to test in `dput()` and the result goes into the test-file.

``` r
library(bayestestR)
dput(map_estimate(rnorm_perfect(1000)))
#> structure(-0.00368006135440124, MAP_density = 0.389791625903652, class = c("numeric", 
#> "MAP"))
```

DominiqueMakowski: I hope this attribute won't create problems for other use cases apart from testing 🤞

strengejacke: No, except you compare objects with `all.equal()` somewhere ;-) Else, attributes are pretty harmless. Even such harmless, that they're sometimes simply forgotten and dropped (e.g. after using `base::subset()`).

strengejacke: Every class-_attribute_ is an _attribute_ as well. :-D

DominiqueMakowski:
Seems like travis testing fails *again*, not sure why

```
8.28s$ Rscript -e 'deps <- devtools::dev_package_deps(dependencies = NA);devtools::install_deps(dependencies = TRUE);if (!all(deps$package %in% installed.packages())) { message("missing: ", paste(setdiff(deps$package, installed.packages()), collapse=", ")); q(status = 1, save = "no")}'
Error in match.arg(upgrade, c("ask", "always", "never")) : 
  'arg' must be of length 1
Calls: <Anonymous> ... upgradable_packages -> resolve_upgrade -> match.arg
Execution halted
```

Issue #77, opened 2019-04-26
Add function for Savage-Dickey density ratio Bayes factors


mattansb:
Would you consider adding a function for a Savage-Dickey density ratio Bayes factor?

Something like this, perhaps?

````
savage_dickey_BF <- function(xpost,xprior,null = 0,direction = 0){
  if (!require(logspline)) {
    stop('Need logspline to run')
  }
  f_post <- suppressWarnings(logspline(xpost))
  f_prior <- suppressWarnings(logspline(xprior))
  
  d_post <- dlogspline(null,f_post)
  d_prior <- dlogspline(null,f_prior)
  
  norm_post <- norm_prior <- 1
  if (direction < 0) {
    norm_post <- plogspline(null,f_post)
    norm_prior <- plogspline(null,f_prior)
  } else if (direction > 0) {
    norm_post <- 1-plogspline(null,f_post)
    norm_prior <- 1-plogspline(null,f_prior)
  }
  
  (d_post/norm_post)/(d_prior/norm_prior)
}
````

strengejacke:
Do you have any references on this approach / method?

DominiqueMakowski:
I am currently reading [this](https://www.ejwagenmakers.com/2010/WagenmakersEtAlCogPsy2010.pdf) and [this](https://michael-franke.github.io/statistics,/modeling/2017/07/07/BF_computation.html). I wasn't aware of this approach ☺️

mattansb:
Seems like @DominiqueMakowski found the Wagenmakers reference ☺️

DominiqueMakowski:
And also [Heck 2018](https://www.ncbi.nlm.nih.gov/pubmed/30451277)  for a "generalized" version in regards to the strong assumptions of this approach.

It *seems* to me that this index is used more in the context of model performance than parameters (coefficients) testing. Thus, I wonder whether we should add it here or in [performance](https://github.com/easystats/performance) where R2_bayes is currently implemented.

mattansb:
IMO it should be here, as it goes very well with your other [NHST](https://github.com/easystats/bayestestR#null-hypothesis-significance-testing-nhst) functions.

Note also that `brms` uses this method in the `hypothesis` function for testing parameters and parameter-combinations.

DominiqueMakowski:
[...]

PR #79, opened 2019-04-26
new bayes_factor functions

mattansb:
as per #77 discussion, added the `bayes_factor` (etc) function.
Also added a suggestion for `logspline` function as it produces more robust Savage-Dickey BFs.


mattansb:
`bayes_factor` and `bf` might need a name change to avoid masking / being masked by `brms::bayes_factor.brmsfit`.

Maybe `Bayes_factor` / `Bf` ?

R/bayes_factor.R
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  return(bf_val)
+}
+
+#' @rdname bayes_factor
+#' @export
+#' @import logspline
+savage_dickey_bf <- function(x,prior,null = 0,direction = 0){
+  if (require(logspline)) {
```
strengejacke: I would change this into `requireNamespace()`, like

```r
if (!requireNamespace("logspline")) {
  stop("Package 'logspline' required.")
}
```

mattansb: Oops! That's what I meant to do!
Changed.

mattansb: Sorry, @strengejacke , you think this function should not run without `logspline`?
Note that the function can return a result using only `bayestestR` / `base` functions, but these tend to be less robust (though usually only slightly different) than those from `logspline`.

DominiqueMakowski:
For the names, we usually try to keep some consistency with some rules, such as:

- lowercase for functions and arguments
- maximum readability (the easystats code and functions should be understandable by my grandma)
- self-organisation using a conjunction of prefix_sufix if applicable

In this context, I would tend to have `bayesfactor` as the "main" function and then `bayesfactor_savagedickey`. TBH a few months ago I would have vouched for `bayes_factor_savage_dickey` but @strengejacke  convinced me that too much of separations and spacing is not that good either. As Thanos would say in Avengers, "perfectly balanced, as all things should be".

R/bayes_factor.R
DominiqueMakowski: We indeed try to keep mandatory dependencies to a minimum, which doesn't preclude having them in `suggests` as Daniel suggested. That means that people can install bayestestR without logspline, and use it, but if they want to use the function that uses it, they will be prompted with a message saying that they have to install it in order to use this function. If they already have it installed then no problem.

In general (we have this kind of stopping in several places throughout easystats), we should maybe reformulate the error message in a more informative and user-friendly manner, such as for example:

```r
stop("This function requires the package 'logspline' to be installed. Please install by running 'install.packages(logspline)' and run the function again :)") 
```

strengejacke: I wasn't aware it also works w/o the logspline package. I'd suggest putting the code that does not rely on logspline into the if-block then, instead of the `stop()`. In any case, I think you should use `requireNamespace()` instead of `require()`, as this is what CRAN prefers.

strengejacke: or simply replace `if (require(logspline)) {` by `if (requireNamespace("logspline")) {`

mattansb: Okay, so I added `logspline` in the suggest, and changed `if (require(logspline)) {` by `if (requireNamespace("logspline")) {`, and allowed the function to run without `logspline` with a warning.

mattansb:
I will change the functions name as you suggest. What about `bf` conflicting with `brms::bf` - any ideas?

DominiqueMakowski:
Not really 😕... I think we can try living without the shortcut at first, and see how it goes.

One of the rationales for having function names organized into "function" and "function_specific" (such as `bayesfactor` and `bayesfactor_savagedickey` is that, in this case, people might start type `bayesfac` and the autocompletion would suggest them all the "related" functions and derivatives, which they can they can then learn about. I truly believed that there is many great functions and algorithms existing out there which are almost never used because people just don't know that it exists. So we also try to increase discoverability :)

strengejacke:
Looks good now, I think! I did some minor revisions, mainly:

- code-style, using `Ctrl+Shift+A` (styler) in RStudio, i.e. following the RStudio code-stlye suggestions
- explicit use of namespace (`logspline::`)

mattansb:
Look good!
I added an example in the documentation. I think that's it?

strengejacke:
Yes. Just a minor comment, I think you don't need the output of the example-code as comments in the examples.

@DominiqueMakowski what do you say? I think it's ready to be merged.

DominiqueMakowski:
That's awesome work! A great addition ☺️. I did some very minor changes (for example, added dots at the end of sentences...). Few comments:

1. Renamed `null` argument for `h0`. If I understood correctly (I might have not), this argument is the value that we test against, which is usually the null hypothesis i.e., 0. However, as people might to want to test against other values than the null, wouldn't it be more generic to name it h0 than null? What do you think?
2. Added it in the print (and changed the name of the object returned `bf_val` -> `bayesfactor`)
3. I wonder if there is a way of setting a "default" value for the prior. For instance, a normal distribution located at h0 value with an SD of 1? Or something similar to the t-tests of the BayesFactor package (that uses a Cauchy distribution with a scale of √2/2. Also, maybe the prior characteristics (location and scale?) would be worth displaying in the print (I don't use Bayes factors much so it's a genuine question)?
4. I think it would be worth adding a more thorough documentation (we can extract chunks from the Wagenmakers paper) about what it is. 
5. This expanded documentation could be added to the README with a nice figure helping to visualise it (this could be done in a next update).

DominiqueMakowski:
I've added bayesfactor to the README :)

mattansb:
The doc looks great!

Regarding point 3: I can't think of any default prior that would make sense - we could use a JZS one, but that would only make sense in very specific scenarios, and in general I think it would be better if users *had* to supply their own prior samples.

Also, if we add other methods, the `prior` and `direction` arguments only make sense for `savage-dickey`. That is why I originally added those only to the `bayesfactor_savagedicke` function, and not the "main" `bayesfactor` function. I think this should be reverted back. Do you agree?

DominiqueMakowski:
> Regarding point 3: I can't think of any default prior that would make sense - we could use a JZS one, but that would only make sense in very specific scenarios, and in general I think it would be better if users had to supply their own prior samples.

Makes sense. It would be nice though to document prior selection with recommended and/or commonly used priors. To orient the user.


> Also, if we add other methods, the prior and direction arguments only make sense for savage-dickey. That is why I originally added those only to the bayesfactor_savagedicke function, and not the "main" bayesfactor function. I think this should be reverted back. Do you agree?

I see! It's just that usually the main method definition includes all the possible parameters (it's then easy to see the defaults values and possibilities when the user calls for help on the method). It's mainly a matter of consistency. Another possibility, similar to what is done for instance in [report](https://github.com/easystats/report/blob/f5efe7a2b1e48d4353bbb8f9994c87e6d4e02931/R/report.R#L3), is to have a separate documentation for Savage-Dickey's method and link it in the general method. @strengejacke what are your thoughts on that?

mattansb:
Actually, now that I think about it, the naming of `H0` might also be confusing to the user, as technically for  savage-dickey, H0 (what we're testing against) is = (0|Prior), so both the prior and the point value.
Perhaps the `prior` should be changed to `H0` and `null` be kept as `null`. This is more aligned with the framing in the Wagenmakers paper.

strengejacke:
The "main method" should be the most generic one, while class specific methods gain extra arguments (like in _insight_ `get_data()`). However, here it's a bit different, because we have different arguments based on the method. I would say we include all required arguments in the main definitions. Especially if we don't have default values, because else it leads to unexpected behaviour. I tried `bayesfactor(posterior)` and got an error because `prior` was missing, but the function docs did not mentioned that argument in the function definition.

In short, I would add all necessary arguments to the function definition and write in the docs which argument is required for which method.

DominiqueMakowski:
About the names of h0 and prior...I might be biased because of my no-familiarity with this method. But it seems more intuitive to keep "prior" to be a distribution (we intuitively expect priors and posteriors to be distributions). I understand that H0 is ("point-value"|prior), but if prior is prior, then it's not that crazy to name the point-value h0 (but indeed we should document that h0 we're testing against is actually a combination between this and the prior). I was just suggesting something in that line because it sounded familiar with the frequentist framework. `null` could be okay, it's just that I feel like it would be weird to have null that is non-zero. And I don't know it looks weird to have an argument named null... But that's only me ☺️. Maybe something like `hypothesis` (although that's close to h0) or `mu`.

Speaking of t-tests, maybe the `direction` argument could accept strings? so that the meaning could be more explicit and more consistent with base functions such as [t-tests](https://www.rdocumentation.org/packages/stats/versions/3.5.3/topics/t.test), which accepts "two.sided" (default), "greater" or "less".

strengejacke:
I would also love to have this for models (brmsfit, stanreg), but not sure what "prior" would be? A distribution around zero and test against zero effect?
Similar to what is suggested here:
https://github.com/easystats/bayestestR/issues/80
but just for one model object.

DominiqueMakowski:
For models, I guess it needs another "model" to test against. However, we could potentially think of a "default" such as a constant model (with just the intercept)?

strengejacke:
Ok, then we have `equivalence_test()` to test against "null", and then it indeed makes sense to have something like `compare_models()` or so for bayesfactor comparison.

DominiqueMakowski:
Maybe `alternative` is a good argument name instead of `h0`. Although usually `h1` is the "alternative" to the null hypothesis, but still... it's understandable that we test a posterior against an alternative value.

mattansb:
I think we should continue this methods/args/function-names discussion in #80 .

DominiqueMakowski:
@mattansb Could you add yourself in the package as author ☺️ ? Then I'll merge this PR and we can polish it further on the dev branch.

mattansb:
@DominiqueMakowski done!

DominiqueMakowski:
@mattansb do you have some tests we could re-use to have this code tested?

mattansb:
I'll add something simple.

mattansb:
@DominiqueMakowski should I add it into dev? to master?

strengejacke:
directly master

DominiqueMakowski:
(if it works, otherwise dev, I don't like having the travis build not passing on master, must have some sort of OCD ^^)

mattansb:
I've cloned `easystats/bayestestR` on my desktop, and tried pushing my commit, and got a error saying I don't have editing permissions :/
Am I doing something wrong? (don't have much experience collaboration on github..)

strengejacke:
> if it works

I somehow can't really get used to work with dev-branches. ;-) I think larger implementations can even get their own branch (besides dev), in most cases I commit to master or every now and then I merge into master, because I else lose track of all the changes done, and get lost where I should actually work (master? dev?) :-D

DominiqueMakowski:
> I've cloned easystats/bayestestR on my desktop, and tried pushing my commit, and got a error saying I don't have editing permissions :/ Am I doing something wrong? (don't have much experience collaboration on github..)

I think you have to open a PR to push to the master, there is no "automatic" commit directly to the master. Hence, pushing to the dev branch 😬 can work since it's usually always piped to the master with a PR.

mattansb:
Hmm nope - can't push to dev either :/
I also can't edit any files in the web interface of `easystats/bayestestR`, so it seems like I really am lacking some permission...

DominiqueMakowski:
Should be ok now.

Issue #70, opened 2019-04-15
HDI_area attribute as data frame also for rope.numeric()


strengejacke:
``` r
model <- rstanarm::stan_glm(mpg ~ wt + cyl, data = mtcars)
attributes(rope(model))$HDI_area
#> $`(Intercept)`
#>     CI  HDI_low HDI_high
#> 1 0.90 36.60966 42.66055
#> 2 0.95 36.12163 43.45867
#> 
#> $wt
#>     CI   HDI_low  HDI_high
#> 1 0.90 -4.432082 -1.776227
#> 2 0.95 -4.668768 -1.481225
#> 
#> $cyl
#>     CI   HDI_low   HDI_high
#> 1 0.90 -2.254774 -0.8010249
#> 2 0.95 -2.396010 -0.6578996
```

Since we return one or multiple dataframes for models (which I like), could we also return such dataframe for the cases where a vector is passed to ROPE?

``` r
library(bayestestR)
attributes(rope(x = rnorm(1000, 1, 1)))$HDI_area
#> [[1]]
#> [1] -0.5071979  2.7443272
attributes(rope(x = rnorm(1000, 1, 1), ci = c(.90, .95)))$HDI_area
#> [[1]]
#> [1] -0.8034466  2.5204936
#> 
#> [[2]]
#> [1] -0.8451218  2.9656530
```

So the output would like:
``` r
attributes(rope(x = rnorm(1000, 1, 1), ci = c(.90, .95)))$HDI_area
#>     CI  HDI_low HDI_high
#> 1 0.90 36.60966 42.66055
#> 2 0.95 36.12163 43.45867
```
Sorry to bother with this, I would gladly change it myself but I am afraid of breaking everything since the attribute code is still not perfectly clear to me (I am studying it tho :)

_Originally posted by @DominiqueMakowski in https://github.com/easystats/bayestestR/issues/69#issuecomment-483217422_

PR #81, opened 2019-04-27
Update

DominiqueMakowski:


DominiqueMakowski:
@strengejacke tried to fix #70 but broke rope for models... The error seems to be in the big lapply function. But weird, as It is not supposed to use rope.numeric (where I modified the attributes). Any ideas?

DominiqueMakowski:
Good question, although it makes sense since in the end both hdi and quantiles are used as CIs. Otherwise, we would need to have column names HDI_95 and Percentile_95 which is less cool IMO than the unified CI...

DominiqueMakowski:
@strengejacke Unfortunately the tests do not pass:

```r
equivalence_test(rnorm_perfect(1000, 0, 1))
```
```
Error in `$<-.data.frame`(`*tmp*`, "HDI_low", value = c(CI = 95, CI_low = -1.96037520326988,  : 
  replacement has 3 rows, data has 1
```

PR #85, opened 2019-04-30
More BF stuff

strengejacke:


R/bayesfactor_inclusion.R
```diff
@@ -0,0 +1,199 @@
...
+
+    for (eff in effnames) {
+      df.interaction[,eff] <- sapply(effnames, function(x) includes_interaction(x,eff))
+    }
+    rownames(df.interaction) <- effnames
+    df.interaction <- df.interaction[,-1]
+  }
+
+  # Build Effect Table #
+  df.effect <- data.frame(effnames,
```
strengejacke: effnames might be a factor, unless you set `stringsAsFactors = FALSE` here. Could you check if this is working as intended?

R/bayesfactor_inclusion.R
```diff
@@ -0,0 +1,204 @@
...
+      rand_print[[i]] <- paste0("(", tmp_trms, "|", random_units[[i]], ")")
+    }
+
+    stats::setNames(
+      c(fix_trms,unlist(rand_trms)),
+      c(fix_trms,unlist(rand_print))
+    )
+  }
+
+  print_terms <- character()
```
strengejacke: `print_terms` is defined but not used...

R/bayesfactor_inclusion.R
mattansb: Why oh why is `stringsAsFactors = TRUE` by default?

mattansb: It works any way, but I set `stringsAsFactors = FALSE` just to be safe.

R/bayesfactor_inclusion.R
mattansb: removed

strengejacke:
Can we have a merge into master once this somehow works? Finetuning can be added later...

mattansb:
All `bayesfactor_*` functions have been tested to work, so I think it's safe to merge them into the master.

strengejacke:
> Pull requests that have a failing status can’t be merged on a phone.

Haha... Ok, must go to my PC now.

PR #88, opened 2019-05-04
Update

DominiqueMakowski:
Added a new version of `p_direction` based on AUC. My idea is that for posteriors with very few samples, estimating the density function and derivating the pd from that could be more sensitive (and accurate?) than taking the ratio of samples. 

I will add these two versions to the [comparison](https://easystats.github.io/bayestestR/articles/indicesExistenceComparison.html) to see how it performs. What do you think?


strengejacke:
What about making both available? Does that make sense? I must take a closer look at the statistics

DominiqueMakowski:
Yes they are, currently, toggled by the `raw=FALSE` argument. However, I wonder if there is a better name for example `auc=TRUE`

strengejacke:
perhaps a `method`-argument? with optione `auc` and another one.

DominiqueMakowski:
true, method is more consistent with our other functions. `method="auc"` and `method="raw"` or "simple"?

strengejacke:
We did not have the cauchy, poisson and t-functions on CRAN, so no need to deprecate.

DominiqueMakowski:
I am checking the vignette. I've simplified some code a little, especially in plots: although a bit uglier, I think it's worth in such tutorials for beginners to keep the code as clear as possible for them so they can really get what is going on at each step ☺️ 

I've mostly commented out the chunks, but if you agree we can remove them ;)

DominiqueMakowski:
BTW there's an issue in the update method:

```
Error in update.bayesfactor_models(comparison, reference = 3) : 
  object 'x' not found
```

When subset is NULL, no x is generated.

DominiqueMakowski:
I also have issues with the BIC approx:

```r
> bayesfactor_models(m1, m2, m3, m4, denominator = 1)
Error in mBIC - mBIC[denominator] : 
  non-numeric argument to binary operator
```

DominiqueMakowski:
There are some issues related to non-updated code, that makes some tests fail and prevent me to complete the vignette.

Nevertheless, it's great! A very good overview of the purpose and capabilities of these features.

DominiqueMakowski:
no probs, but isn't this weird bump closer to reality (in terms of underlying data)?

strengejacke:
@DominiqueMakowski @mattansb 
https://github.com/easystats/bayestestR/pull/88/files#diff-90056688ae5b0932b44b5df2935c1d15L201

Here's a variable `x_subset` which is obviously not used... Not sure, but I guess this should be `object_subset`?

btw. have you enabled diagnostic options in RStudio? That helps spotting such things, also directly checks code-style etc.

image

mattansb:
> Here's a variable x_subset which is obviously not used... Not sure, but I guess this should be object_subset?

Oops, that on me I think. Will fix now (and enabled diagnostic options).

mattansb:
>no probs, but isn't this weird bump closer to reality (in terms of underlying data)?

We're really interested here in the prior and posterior distributions (and the samples are just proxies for this), so that weird bump is just an artifact from the smoothing done by `stats::density` to go from samples - > distribution.

R/print.bayesfactor_models.R
```diff
@@ -1,21 +1,23 @@
...
+  BFE_ <- BFE_[-denominator, 'log.BF', drop = FALSE]
   colnames(BFE_) <- ""
 
   cat('Bayes factor analysis\n--------------')
-  print.data.frame(BFE_,digits = digits)
+  print.data.frame(BFE_, digits = digits)
   cat('\nAgainst denominator:\n\t', denM)
-  cat('\n---\nBayes factor type: ', grid.type)
-  if (log) cat('\nPresenting log(BF)')
-  invisible(x)
```
mattansb: @strengejacke Do we not want to return the original object invisibly?

strengejacke: I don't think this is necessary. If you need the object, you call `x <- bayesfactor_models()`. If the print-method returns an object, you can call `x <- print(bayesfactor_models())` to save the object, so there's no benefit.

mattansb: Edge case: using print as a side-effect:

````
library(bayestestR)
library(magrittr)

lm1 <- lm(Sepal.Length ~ 1, data = iris)
lm2 <- lm(Sepal.Length ~ Species, data = iris)
lm3 <- lm(Sepal.Length ~ Species + Petal.Length, data = iris)
lm4 <- lm(Sepal.Length ~ Species * Petal.Length, data = iris)

bayesfactor_models(lm1, lm2, lm3, lm4, denominator = 1) %>% 
  print() %>%
  bayesfactor_inclusion()
````

Comment (strengejacke) [reply R/print.bayesfactor_models.R]: You can do this like:

```r
bayesfactor_models(lm1, lm2, lm3, lm4, denominator = 1) %T>% 
  print() %>%
  bayesfactor_inclusion()
```

mattansb: What if I only loaded `dplyr`? :P

R/p_direction.R
```diff
@@ -35,28 +40,40 @@ pd <- p_direction
...
+    p_direction <- 100 * max(
+      c(
+        length(x[x > 0]) / length(x), # pd positive
+        length(x[x < 0]) / length(x) # pd negative
+      )
     )
-  )
-
+  } else {
+    dens <- as.data.frame(stats::density(x, n = 2^10))
```
DominiqueMakowski: @mattansb Based on what you mentioned about the difference between logspline and base density, would you recommend using the logspline density instead for density here?

mattansb: `logspline`'s algorithm seems to give more sensible results (especially around distribution tails/edges). So if edges/tails are important for AUC calculations, it might be better to use `logspline`. If not, base is probably fine.

DominiqueMakowski: As the tail is the only thing of importance here, I think it would make sense 😅

DominiqueMakowski: Maybe we should implement our wrapper for logspline density, since it's used at different places. Something like `estimate_density` or something.

DominiqueMakowski: (we could even ~steal~ borrow the [logspline](https://github.com/cran/logspline/blob/master/R/logspline.R) function to avoid the dep issue...

strengejacke:
Can we merge this again in the near future? It always irritates me seeing open issues that have been already solved when not working in master. ;-)

strengejacke:
I also suggest the master-branch for me and the padawan-branch for @DominiqueMakowski  ;-)

DominiqueMakowski:
will merge as soon as it passes travis :smiling_imp:

R/p_direction.R
DominiqueMakowski: Unfortunately, there is some compiled C code in the package that it makes it copy non-trivial. Should we exceptionally add logspline to the dependencies (it is quite light)?

DominiqueMakowski:
`bayesfactor_inclusion` is failing at my place, is that also the case with you?

R/print.bayesfactor_models.R
mattansb: It seems to be standard practice in R for printing methods to return the input object (e.g. `stats`'s `print` functions). I'm adding this back in.

mattansb:
Ran the tests - only getting the expected warning. Where is it failing?

R/p_direction.R
mattansb: If we want to keep dependency light, you can default to `stats::density` method if `logspline` is missing, with a message (like in `bayesfactor_savagedickey`):
````
insight::print_color("Consider installing the \"logspline\" package for a more robust estimate.\n", "red")
````

**Comment by DominiqueMakowski** [comment]
```r
>   ToothGrowth$dose <- as.factor(ToothGrowth$dose)
>   BF_ToothGrowth <- BayesFactor::anovaBF(len ~ dose*supp, ToothGrowth)
  |======================================================================================| 100%
>  bayestestR::bayesfactor_inclusion(BF_ToothGrowth)
```

```
 Error in `*tmp*`[[j]] : recursive indexing failed at level 2 
6.
`[<-.data.frame`(`*tmp*`, m, tmp_terms, value = TRUE) at bayesfactor_inclusion.R#202
5.
`[<-`(`*tmp*`, m, tmp_terms, value = TRUE) at bayesfactor_inclusion.R#202
4.
.get_model_table(models, priorOdds = prior_odds) at bayesfactor_inclusion.R#67
3.
bayesfactor_inclusion.bayesfactor_models(models, match_models = match_models, 
    prior_odds = prior_odds) at bayesfactor_inclusion.R#135
2.
bayesfactor_inclusion.BFBayesFactor(BF_ToothGrowth) at bayesfactor_inclusion.R#59
1.
bayestestR::bayesfactor_inclusion(BF_ToothGrowth) 
```

mattansb:
Same by me:
``` r
ToothGrowth$dose <- as.factor(ToothGrowth$dose)
BF_ToothGrowth <- BayesFactor::anovaBF(len ~ dose*supp, ToothGrowth)
bayestestR::bayesfactor_inclusion(BF_ToothGrowth)
#>           Pr(prior) Pr(posterior) Inclusion.BF
#> supp            0.6          1.00      1.4e+02
#> dose            0.6          1.00      3.2e+14
#> supp:dose       0.2          0.72      1.1e+01
#> ---
#> Inclusion BFs compared among all models.
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

That is very weird...  especially since it seems to fail in `bayesfactor_inclusion.bayesfactor_models`. I might have an idea though - can you try and run this code, and see if you get the same results?

``` r
ToothGrowth$dose <- as.factor(ToothGrowth$dose)
BF_ToothGrowth <- BayesFactor::anovaBF(len ~ dose*supp, ToothGrowth)
bayestestR::bayesfactor_models(BF_ToothGrowth)
#> Bayes factor analysis
#> --------------                                      
#> [2] supp                       1.2e+00
#> [3] dose                       5.0e+12
#> [4] supp + dose                2.9e+14
#> [5] supp + dose + supp:dose    7.8e+14
#> 
#> Against denominator:
#>   [1] (Intercept only)   
#> ---
#> Bayes factor type:  JZS (BayesFactor)
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

DominiqueMakowski:
``` r
ToothGrowth$dose <- as.factor(ToothGrowth$dose)
BF_ToothGrowth <- BayesFactor::anovaBF(len ~ dose*supp, ToothGrowth)
bayestestR::bayesfactor_inclusion(BF_ToothGrowth)
#> Error in `*tmp*`[[j]]: recursive indexing failed at level 2

ToothGrowth$dose <- as.factor(ToothGrowth$dose)
BF_ToothGrowth <- BayesFactor::anovaBF(len ~ dose*supp, ToothGrowth)
bayestestR::bayesfactor_models(BF_ToothGrowth)
#> Bayes factor analysis
#> --------------                                      
#> [2] supp                       1.2e+00
#> [3] dose                       5.0e+12
#> [4] supp + dose                2.8e+14
#> [5] supp + dose + supp:dose    7.6e+14
#> 
#> Against denominator:
#>   [1] (Intercept only)   
#> ---
#> Bayes factor type:  JZS (BayesFactor)
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


Weird indeed... and insight is updated. Let me update BayesFactor

mattansb:
It shouldn't have anything to do with `BayesFactor`, as `bayesfactor_inclusion.BFBayesFactor` calls `bayesfactor_models.BFBayesFactor` (which works fine) and then passes the results to `bayesfactor_inclusion.bayesfactor_models`. :/

Does this work?
``` r
library(bayestestR)
mo0 <- lm(Sepal.Length ~ 1, data = iris)
mo1 <- lm(Sepal.Length ~ Species, data = iris)
mo2 <- lm(Sepal.Length ~ Species + Petal.Length, data = iris)
mo3 <- lm(Sepal.Length ~ Species * Petal.Length, data = iris)

BFmodels <- bayesfactor_models(mo1, mo2, mo3, denominator = mo0)
BFmodels
#> Bayes factor analysis
#> --------------                                     
#> [1] Species                   1.7e+29
#> [2] Species + Petal.Length    5.8e+55
#> [3] Species * Petal.Length    2.2e+54
#> 
#> Against denominator:
#>   [4] (Intercept only)   
#> ---
#> Bayes factor type:  BIC approximation
bayesfactor_inclusion(BFmodels)
#>                      Pr(prior) Pr(posterior) Inclusion.BF
#> Species                   0.75         1.000      2.0e+55
#> Petal.Length              0.50         1.000      3.6e+26
#> Species:Petal.Length      0.25         0.036      1.1e-01
#> ---
#> Inclusion BFs compared among all models.
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

DominiqueMakowski:
``` r
library(bayestestR)
mo0 <- lm(Sepal.Length ~ 1, data = iris)
mo1 <- lm(Sepal.Length ~ Species, data = iris)
mo2 <- lm(Sepal.Length ~ Species + Petal.Length, data = iris)
mo3 <- lm(Sepal.Length ~ Species * Petal.Length, data = iris)

BFmodels <- bayesfactor_models(mo1, mo2, mo3, denominator = mo0)
BFmodels
#> Bayes factor analysis
#> --------------                                     
#> [1] Species                   1.7e+29
#> [2] Species + Petal.Length    5.8e+55
#> [3] Species * Petal.Length    2.2e+54
#> 
#> Against denominator:
#>   [4] (Intercept only)   
#> ---
#> Bayes factor type:  BIC approximation

bayesfactor_inclusion(BFmodels)
#> Error in `*tmp*`[[j]]: subscript out of bounds
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Nope... you have built the latest dev?

mattansb:
> Nope... you have built the latest dev?

Yup...

Looking at your output, the failure seems to happen in [line 202](https://github.com/easystats/bayestestR/blob/dev/R/bayesfactor_inclusion.R#L202), where the terms for each model are marked. But up until that point in the function, only `base` and `stats` functions are used, so assuming `bayesfactor_models` runs without fail (which it does), I can't imagine what the F is going on :/

@strengejacke are you getting the same failure?

mattansb:
I am running `R` 3.5.3... But that couldn't be it, right?

DominiqueMakowski:
I hope @strengejacke can check despite the fact that it's on the dark side of the branches 😉

DominiqueMakowski:
I'll upgrade just in case

DominiqueMakowski:
``` r
library(bayestestR)
mo0 <- lm(Sepal.Length ~ 1, data = iris)
mo1 <- lm(Sepal.Length ~ Species, data = iris)
mo2 <- lm(Sepal.Length ~ Species + Petal.Length, data = iris)
mo3 <- lm(Sepal.Length ~ Species * Petal.Length, data = iris)

BFmodels <- bayesfactor_models(mo1, mo2, mo3, denominator = mo0)
BFmodels
#> Bayes factor analysis
#> --------------                                     
#> [1] Species                   1.7e+29
#> [2] Species + Petal.Length    5.8e+55
#> [3] Species * Petal.Length    2.2e+54
#> 
#> Against denominator:
#>   [4] (Intercept only)   
#> ---
#> Bayes factor type:  BIC approximation

bayesfactor_inclusion(BFmodels)
#>                      Pr(prior) Pr(posterior) Inclusion.BF
#> Species                   0.75         1.000      2.0e+55
#> Petal.Length              0.50         1.000      3.6e+26
#> Species:Petal.Length      0.25         0.036      1.1e-01
#> ---
#> Inclusion BFs compared among all models.
```

<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

... Okay so it seems to work with 3.6.0...

mattansb:
>... Okay so it seems to work with 3.6.0...

🤷‍♂️

DominiqueMakowski:
The fact that it doens't work on all R versions is quite problematic though, and it's unfortunately very hard to test and assess 😕

mattansb:
This whole thing is super weird, because my original `BFEffects::inclusionBF` used the [same exact code](https://github.com/mattansb/BFEffects/blob/master/R/inclusionBF-utility.R#L60), and that been running since June 2018 at least, so on R 3.5.0, if not 3.4.4...

DominiqueMakowski:
even worse is that I recall that I ran without any probs at the beginning...

mattansb:
Might have been something local to your setup? And any new installation of R would have worked?
Want to try to reinstall the previous version of R you had, and see if it works to test this hypothesis?

DominiqueMakowski:
I'll try tomorrow at my uni's pc which has a different version, we'll see :)

strengejacke:
Have you guys already discovered the debugging-feature in RStudio, or are you continuously using reprexes to debug your code? :-P

DominiqueMakowski:
@strengejacke Do you know that I've been trying to understand debugging for R and python for like 4 years... And yet I still print(1), print(2) and print(3) 😅 

You mean using pointbreaks and stuff?

strengejacke:
> pointbreaks

Probably, it's called breakpoints... Not suprising it already took 4 years! :-D

DominiqueMakowski:
sometimes like Yoda I speak 😁

strengejacke:
The problem is inside the sub-function `make_terms()`:

```r
    if (length(random_parts) == 0) {
      return(stats::setNames(fix_trms, fix_trms))
    }
```

We have:

```r
as.data.frame(BFmodels)
#>                    Model    log.BF
#> 1                Species  67.30315
#> 2 Species + Petal.Length 128.40744
#> 3 Species * Petal.Length 125.12953
#> 4                      1   0.00000
```

Rows are reorder, so `make_terms()` is called with `"1"` in the first instance:

```r
  for (m in seq_len(nrow(df.model))) {
    tmp_terms <- make_terms(df.model$Modelnames[m])
    df.model[m, tmp_terms] <- TRUE
  }
```

Since the "1" has no term labels when coerced to fomula., `all.terms` is empty, and so is `fix_trms`.

```r
make_term <- function(formula) {
    formula.f <- stats::as.formula(paste0('~', formula))
    all.terms <- attr(stats::terms(formula.f), "term.labels")

    fix_trms <- all.terms[!grepl("\\|", all.terms)] # no random

    random_parts <- paste0(all.terms[grepl("\\|", all.terms)]) # only random
    if (length(random_parts) == 0) {
      return(stats::setNames(fix_trms, fix_trms))
    }
...
```

In the next step of the for-loop, you assign `TRUE` to a column that doesn't exist, because `tmp_terms` is an empty character.

```r
df.model[m, tmp_terms] <- TRUE
```

This works on R3.6, but not on R < 3.6 (just tested here).

But: I don't understand the whole code. This loop 

```r
  for (m in seq_len(nrow(df.model))) {
    tmp_terms <- make_terms(df.model$Modelnames[m])
    df.model[m, tmp_terms] <- TRUE
  }
```

generates _column names_ from terms, but these _columns_ do not exist in the `bayesfactor_models`-object (the above data frame). 

What is the purpose of that loop and the `make_terms`-function? I think we can easily replace that code in `make_terms()` with a lot shorter code by some _insight_-function. And then it needs to be checked what the for-loop actually should do, as those columns never exist, so `.get_model_table()` always ends up in returning an empty or invalid data frame.

Also, `df.model[is.na(df.model)] <- FALSE` is probably inaccutate.

mattansb:
Wow, good eye @strengejacke ! (Stupid intercept-only models...) I just pushed a fix for this (note that it works fine for me on R 3.5.3, and it did even before that.)

---

`.get_model_table` returns the following table, that has all the information needed to compute inclusion BFs.

``` r
library(bayestestR)
mo0 <- lm(Sepal.Length ~ 1, data = iris)
mo1 <- lm(Sepal.Length ~ Species, data = iris)
mo2 <- lm(Sepal.Length ~ Species + Petal.Length, data = iris)
mo3 <- lm(Sepal.Length ~ Species * Petal.Length, data = iris)

BFmodels <- bayesfactor_models(mo1, mo2, mo3, denominator = mo0)

bayestestR:::.get_model_table(BFmodels, priorOdds = NULL)
#>               Modelnames priorProbs    postProbs Species Petal.Length Species:Petal.Length
#> 1                      1       0.25 1.649232e-56   FALSE        FALSE                FALSE
#> 2                Species       0.25 2.796852e-27    TRUE        FALSE                FALSE
#> 3 Species + Petal.Length       0.25 9.636634e-01    TRUE         TRUE                FALSE
#> 4 Species * Petal.Length       0.25 3.633659e-02    TRUE         TRUE                 TRUE
```
<sup>Created on 2019-05-06 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

What the internal `make_term` does is it breaks down a model formula (kind of) into its terms, including random effects:

````
make_terms("1")
#> character(0)
make_terms("0")
#> character(0)
make_terms("A")
#> "A" 
make_terms("A + (1|ID)")
#>    "A" "1:ID" 
make_terms("A + (A|ID)")
#>    "A" "1:ID" "A:ID" 
make_terms("0 + A + (A|ID)")
#>    "A" "1:ID" "A:ID" 
````

This is how the df above is built.

**Comment by DominiqueMakowski** [comment]
The current version passes all checks and all at my place ☺️ 

Just need to please lord travis now: 
```
  ══ testthat results  ═══════════════════════════════════════════════════════════
  OK: 65 SKIPPED: 4 WARNINGS: 36 FAILED: 3
  1. Error: (unknown) (@test-bayesfactor.R#40) 
  2. Error: brms (@test-brms.R#14) 
  3. Error: rstanarm (@test-rstanarm.R#31)
```
Either by skipping these tests on travis if we don't find any other way.

mattansb:
The only Travis I know is Travis Barker... I don't think he'll be much help...

DominiqueMakowski:
Unfortunately, [this travis](https://travis-ci.org/easystats/bayestestR/builds/529139664#L1149) is the one that rhythms our lives now 😥 🥁

DominiqueMakowski:
merging 🎉 Thanks for this long but fruitful exchange

PR #96, opened 2019-05-09
update

DominiqueMakowski:
Mainly polishing:

- [ ] Find why testing fails
- [ ] Improve testing 
- [ ] ...


DominiqueMakowski:
What do you think of the introduction of the BF vignette, that simply present and explicit the notion that it seems to be two distinct schools of thought, which can be roughly summarised as the pro-BF and the anti-BF. I think something along those lines (please do rephrase or modify) can help the user understand where the BFs, or other indices, fall in the Bayesian landscape. What are your thoughts?

mattansb:
@DominiqueMakowski Looks really excellent and clear! I made some minor changes only.
(Only problem I have is that you systematically changed all my american "center"s to the UK "centre"s!)

As you can see, I've also added `bayesfactor_savagedickey.stanreg` to the vignette (and example doc, and tests). It works as expected!

DominiqueMakowski:
@mattansb my computer's spellcheck is biased toward UK 😬

DominiqueMakowski:
Let's merge this, even though I didn't improve the tests coverage as I wanted

PR #105, opened 2019-05-12
update

DominiqueMakowski:


DominiqueMakowski:
@mattansb is that what you meant?

mattansb:
Exactly!

DominiqueMakowski:
Well, although I thought it was not for today, I somehow completed the Example 1 vignette and took in a somewhat different direction than what I was thinking about originally (I was planning just to show the high-level functions to get everything, not make this kind of in-depth step-by-step tutorial). 

Anyway, please take a look at this and the other vignettes, tell me what you think, and feel free to change as there must be plenty of typos/errors and rephrasings to do 😉

DominiqueMakowski:
Ok I'll merge so the vignettes are on the website

mattansb:
It looks really cool! I've already added this as a link to me teaching
materials!

I will add a sdBF to your example later today or tomorrow (I feel like
we're on opposing teams in the posterior / BF argument... 😅)

--
Mattan S. Ben-Shachar, PhD student
Department of Psychology & Zlotowski Center for Neuroscience
Ben-Gurion University of the Negev
The Developmental ERP Lab

DominiqueMakowski:
Haha, well there are actual scientific reasons to be on our side though... 😁 

I would hold on a bit for the sdBF in this example. I would rather see it mentioned in example 2 or such, at a bit more advanced stage. I've included pd and ROPE in example 1 because their understanding and computation is very very straightforward, and tied to the posterior distribution (which is the focus of example 1).

(Also, because these are the two only indices we need, based on [preliminary data](https://easystats.github.io/bayestestR/articles/indicesExistenceComparison.html) 😬😅)

((but don't worry, it's never too late to convert to the anti-BFs, we are very forgiving 😉))

PR #113, opened 2019-05-13
update

DominiqueMakowski:


DominiqueMakowski:
``` r
library(bayestestR)
as.data.frame(bayesfactor_savagedickey(rnorm(1000, 1), prior=rnorm(1000)))
#> Loading required namespace: logspline
#>       BFsd
#> 1 1.677963
```

<sup>Created on 2019-05-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

The current column name is `BFsd`. Could we rename it to `BF` so it will look sleeker in the posterior description.

PR #122, opened 2019-05-15
Dev

strengejacke:


tests/testthat/test-bayesfactor.R
```diff
@@ -25,11 +25,12 @@ test_that("bayesfactor_savagedickey", {
   bfsd <- bayestestR::bayesfactor_savagedickey(model)
   testthat::expect_equal(log(bfsd$BF), c(-2.69, -0.14), tolerance = 0.1)
 
-  library(brms)
-  brms_mixed_6 <- circus::download_model("brms_mixed_6")
-  set.seed(222)
-  sdBF <- bayesfactor_savagedickey(brms_mixed_6, effects = "fixed")
-  testthat::expect_equal(log(sdBF$BF), c(-6.0, -5.8, 0.7, -2.7, -7.4), tolerance = 0.1)
+  # SKIP FOR TRAVIS
```
mattansb: I really am starting not to like this Travis guy...

DominiqueMakowski: Yeah he can be hard to please 😅

DominiqueMakowski:
We should refrain from using `see` before it makes it to CRAN, otherwise, we won't be able to publish this. So the priority is to push `see`, and then update bayestestR on CRAN, and then push `parameters` (which depends on bayestestR).

mattansb:
Sorry, not sure I understood you - should I roll back the plotting change on the vignette then?

DominiqueMakowski:
> should I roll back the plotting change on the vignette then?

If they rely on `see`, yes otherwise it won't be built on CRAN (and thus they won't accept it). However it's best not to roll back, but just comment out the stuff related to plotting for now. 

Then, once `see` is out, we will start documenting and writing about the visualisation options :) (and we will uncomment these paragraphs)

mattansb:
The only part that relies of `see` is the visualization of prior and posterior in the BF vignette.
I will comment this out and add back the previous plotting code.

strengejacke:
Is there really a need to comment out / roll back? bayestestR relies on _see_ anyway, we have started implemented _see_ as if it's available on CRAN, so we need publish see on CRAN anyway. And we don't stop developing bayestestR in parallel as well... _see_ not only used in the vignettes, but also in the many plot-methods already.

mattansb:
So I'm uncommenting out the commented out stuff again (:

DominiqueMakowski:
😬

strengejacke:
> Oh you mean that we should release see before updating bayestestR v2 anyway.

Yes, that's how I understood our effort the last days when we re-structured the plot-methods :-)

DominiqueMakowski:
yes yes absolutely. 

You conditioned me to react negatively each time we depend on unpublished remotes 😁

PR #125, opened 2019-05-16
Dev

DominiqueMakowski:


strengejacke:
for _insight_, we have some obvioisly simple tests that don't pass on Linux. A common pattern seems to be those functions are failing that try to access the model data frame the environment. Not sure if this is different on the different OS?

DominiqueMakowski:
good question...

DominiqueMakowski:
It passes on [everything](https://travis-ci.org/easystats/bayestestR/builds/533378638?utm_source=github_status&utm_medium=notification) BUT linux devel, where it terminated just because the time to run was too long (> 50 min). I believe this is related not to bayestestR directly but to lower deps such as rstan. Disabling linux again for now.

strengejacke:
Or make some of the tests conditional, not running on Travis.

DominiqueMakowski:
But then our codecoverage won't be accurate. So I think it's best to test it as we do currently on all 3 versions, on macOS (so we have a better view of the testing coverage). Eventually from time to time we will check to see if linux still has that minor issue in its devel version.

strengejacke:
Ok. I think it's not absolutely necessary to test on all platforms anyway. But for CRAN we must see not to have to long running tests

DominiqueMakowski:
Yes but I don't think it's our tests that are the source of the issue. It seems like for some reason linux devel has troubles finding and installing rstan... Anyway, it's a minor issue, mainly meant to fulfil my testing-related OCD 😁

vignettes/bayes_factors.Rmd
```diff
@@ -21,7 +21,7 @@ bibliography: bibliography.bib
 
 This vignette can be referred to by citing the package:
 
-- Makowski, D., Ben-Shachar M. S. \& L?�decke, D. (2019). *Understand and Describe Bayesian Models and Posterior Distributions using bayestestR*. Available from https://github.com/easystats/bayestestR. DOI: [10.5281/zenodo.2556486](https://zenodo.org/record/2556486).
+- Makowski, D., Ben-Shachar M. S. \& Lüdecke, D. (2019). *Understand and Describe Bayesian Models and Posterior Distributions using bayestestR*. Available from https://github.com/easystats/bayestestR. DOI: [10.5281/zenodo.2556486](https://zenodo.org/record/2556486).
```
DominiqueMakowski: This can only mean one thing... someone opened the file with A MAC 😠

DominiqueMakowski:
A little help is welcome to add `diagnostic_posterior` for brms ☺️

strengejacke:
will add a mcse-method later, than we have the basis for `diagnostic_posterior.brmsfit()`. I'll address this today.

Issue #144, opened 2019-05-21
ROPE sensitivity to scales?


mattansb:
From [ROPE article](https://easystats.github.io/bayestestR/articles/region_of_practical_equivalence.html):
> However, keep in mind that unlike indices of effect existence (such as the pd), indices of effect size, such as the ROPE percentage, depend on the unit of its parameter, and can thus be easily changed via the scaling of the predictors.

1. I'm far from a ROPE expert, but shouldn't %ROPE be in theory not sensitive to scale? Shouldn't the doc suggest that the ROPE range be adjusted according the the IV's scale?

2. Also, I'm not sure how %ROPE is a measure of effect size - it only indicated the degree by which the effect is not small, but give no indication of the size the effect (if it is medium, or large). 

(I've only now started reading Kruschke's book, after reading some of his papers, so sorry if these are dumb Qs)

mattansb:
1. I guess my question is: if the ROPE range is adjusted for the scale of the DV, why not for the IV? (Maybe this can also be implemented with something like what is used for `parameters::standardize_parameters(x, method = "full")`?

strengejacke:
@1 The rope range is not _adjusted_ for the range of the DV, it is _based_ on that range (at least for linear models). If you standardize your predictors, you'll change the results from the rope, as the rope range does not change (since it's based on the unchanged DV), but the posterior from the scaled IV changed. So "everything working as intended". For p_direction, it doesn't matter if an IV is scaled or not.

strengejacke:
> Also, I'm not sure how %ROPE is a measure of effect size 

True, maybe we should revise the wording here?

DominiqueMakowski:
> Also, I'm not sure how %ROPE is a measure of effect size

I might have been unclear, let me try to rephrase:

the rope percentage is an index of significance (in the sense of "practical importance", rather than some esoteric and conceptual definition), thus it tries to discriminate between significant and non-significant effects. It is not an index of effect size per se. However it uses effect size to define what significance is. It defines a region assimilable to null (or negligible) effect, based solely on the effect's magnitude (I.e., size). Hence it is not directly an index of effect size, but it uses effect size to define significance. Due to this close relationship, both the arguments for and against the focus on effect size apply.

what do you think? 

but yes that is worth reformulating and detailing in the docs.

mattansb:
If I understand the both of you, the %ROPE is a measure of significance, whose criterion is based on effect size.

[...]

PR #149, opened 2019-05-23
Dev

DominiqueMakowski:


R/rope.R
```diff
@@ -10,7 +10,7 @@
 #'
 #' @details Statistically, the probability of a posterior distribution of being
 #'   different from 0 does not make much sense (the probability of it being
-#'   different from a single point being infinite). Therefore, the idea
+#'   different from a single point being 100%). Therefore, the idea
```
DominiqueMakowski: I think infinite is correct here

vignettes/region_of_practical_equivalence.Rmd
```diff
@@ -72,7 +72,7 @@ If the **full ROPE** is used (*i.e.*, 100\% of the HDI), then the null hypothesi
 
 # Sensitivity to the parameter's scale
 
-However, keep in mind that unlike indices of effect *existence* (such as the [`pd`](https://easystats.github.io/bayestestR/articles/probability_of_direction.html)), indices of effect *size*, such as the **ROPE percentage**, depend on the unit of its parameter, and can thus be easily changed *via* the scaling of the predictors.
+However, keep in mind that unlike other indices of effect **existence** (such as the [`pd`](https://easystats.github.io/bayestestR/articles/probability_of_direction.html)), **ROPE percentage** depend on the unit of its parameter, thus correct interpretation of the ROPE as representing a region of practical equivalence to zero is dependent on the scaling of the predictors.
```
DominiqueMakowski: In this case, maybe `unlike indices of effect *existence* the pd, indices of effect *significance*, such as the ROPE` so it the distinction is clearer

vignettes/region_of_practical_equivalence.Rmd
```diff
@@ -120,7 +120,7 @@ p_direction(model)
 rope(model, ci=1)
 ```
 
-As you can see, by simply dividing the predictory by 100, we **drastically** changed (*we divided it by 100*) the conclusion related to the **percentage in ROPE** (which became very close to `0`): the effect could now be **interpreted as being significant**. Thus, we recommend, when reporting or reading ROPE results, to pay close attention to the unit of the predictors.
+As you can see, by simply dividing the predictor by 100, we **drastically** changed the conclusion related to the **percentage in ROPE** (which became very close to `0`): the effect could now be **interpreted as being significant**. Thus, we recommend paying close attention to the unit of the predictors when selecting the ROPE range (e.g., what coefficient would correspond to a small effect?), and when reporting or reading ROPE results.
```
DominiqueMakowski: in this case, it is not particularly the ROPE range (which is fixed in both cases and depends on the unit of the outcome) but the unit of the predictor

R/rope.R
mattansb: The probability of a single point is 0, so the complementary probability to that is 100-0=100%. No?

vignettes/region_of_practical_equivalence.Rmd
mattansb: Oh yes! I like that! Maybe even something about clinical significance?

vignettes/region_of_practical_equivalence.Rmd
mattansb: Yes, that's kind of the point I think I'm trying to make - the ROPE range should correspond to the predictors' units. How would you change this?

strengejacke: Not sure, but maybe my quote here helps: https://github.com/easystats/bayestestR/issues/144#issuecomment-495274448

R/rope.R
strengejacke: No, I don't think so, I think infinite is correct, or maybe undefined, but I think infinite probability is common in maths.

mattansb:
I've heard of negative probability, but couldn't find any infinite
probability on Google 🤔

DominiqueMakowski:
> I've heard of negative probability, but couldn't find any infinite probability on Google 🤔

Let's just say "the probability of a distribution of continuous values against a single point makes no sense" 😅

mattansb:
Or maybe we can simply say: probability of a single point in a continuous
distribution of values is 0 / undefined?

DominiqueMakowski:
"the probability of a single null value in a continuous distribution is 0"

strengejacke:
Is it 0 or 1/Inf?

strengejacke:
I remember something with Limes from math classes in school...

DominiqueMakowski:
The probability of a single point is 0, therefore the probability of the other values (of the posterior distribution) being different from 0 (against a single null value of probability 0) is 1/0 = 💥 

I might be mistaken tho 😅

mattansb:
Probability in a continuous distribution is the area under the curve between two points - the area between one point, is 0 (equal to the integral - so the change at a single point is 1/Inf, so either is correct, but I think the 0 is easier to understand?)

I think for the probability of the other values you need to subtract, not divide: 100-0 (or 100-1/Inf)=100%

DominiqueMakowski:
I think "the probability of a null hypothesis single value in a continuous distribution is 0" is relatively clear

PR #169, opened 2019-06-06
minor

strengejacke:


strengejacke:
The "resolve conflict" doesn't load, I'll use the master branch to directly commit your fix, ok?

PR #182, opened 2019-06-19
Example 1 edit/proofing

humanfactors:
Apologies for the lack of isolation - it's much harder to commit in chunks when getting in the flow of editing writing. I'm going to work out a better workflow with this over time. Please also let me know if this is the appropriate workflow to get this onto the extant PR stack into master?

**Overall the commit is largely just stylistic and grammatical changes:** ✍✔

* I've adjusted a number of sentences across the example to read more clearly. 👁‍🗨
* Any explicit reference to variables now is quoted in code block tags rather than bolding to make more clear it's part of the R code.
* I've change the plot colors to be a little more readable on the eyes (and improved for color blindness)
* Added in blank lines between code blocks for source readability
* Capitalization of Sunflower 🌻🌻🌻

**General Points**
- I'm not a huge fan of the use of bolding throughout, as I think the emphasis is actually lost with excessive emphasis. I've sparingly removed some of these.
- I've tried to ensure that when a variable name is referred to in text, it's in verbatim code tags.

Sorry, I'm confused by the presence of 13e920c27827bb60656d50d15b3031c07c37d942 in the PR as that was my fork updating to latest dev.

### DIFF
--- a/vignettes/example1.Rmd
+++ b/vignettes/example1.Rmd
@@ -104,7 +104,7 @@ posteriors <- insight::get_parameters(model)
 head(posteriors)  # Show the first 6 rows
 ```
 
-As we can see, the parameters take the form of a wide dataframe with two columns, corresponding to the intercept and the effect of `Petal.Length`. These columns contain the **posterior distributions** of these two parameters. In simple terms, the posterior distribution is a set of different plausible values for each parameter.
+As we can see, the parameters take the form of a lengthy dataframe with two columns, corresponding to the `intercept` and the effect of `Petal.Length`. These columns contain the **posterior distributions** of these two parameters. In simple terms, the posterior distribution is a set of different plausible values for each parameter.
 
 #### About posterior draws
 
@@ -147,7 +147,7 @@ This distribution represents the [probability](https://en.wikipedia.org/wiki/Pro
 
 > **Congrats! You've just described your posterior distribution.**
 
-And this is at the heart of Bayesian analysis. We don't need *p* values, *t* values or degrees of freedom: **everything is there**, within this posterior distribution.
+And this is at the heart of Bayesian analysis. We don't need *p*-values, *t*-values or degrees of freedom: **everything is there**, within this posterior distribution.
 
 Our description above is consistent with the values obtained from the frequentist regression (which resulted in a beta of **0.41**). This is reassuring! Indeed, **in most cases a Bayesian analysis does not drastically change the results** or their interpretation. Rather, it makes the results more interpretable and intuitive, and eaasier to understand and describe.
 
@@ -159,7 +159,7 @@ Unfortunately, it is often not practical to report the whole posterior distribut
 
 1. A **point-estimate** which is a one-value summary (similar to the *beta* in frequentist regressions).
 2. A **credible interval** representing the associated uncertainty.
-3. And some **indices of significance**, giving information about the relative importance of this effect.
+3. Some **indices of significance**, giving information about the relative importance of this effect.
 
 
 #### Point-estimate
--- a/vignettes/example1.Rmd
+++ b/vignettes/example1.Rmd
@@ -48,21 +48,20 @@ library(insight)
 
 ## Simple linear model (*aka* a regression)
 
-We will begin by a simple regression to test the relationship between **Petal.Length** (our predictor, or *independent*, variable) and **Sepal.Length** (our response, or *dependent*, variable) from the [`iris`](https://en.wikipedia.org/wiki/Iris_flower_data_set) dataset, included by default in R. 
+We will begin by conducting a simple linear regression to test the relationship between `Petal.Length` (our predictor, or *independent*, variable) and `Sepal.Length` (our response, or *dependent*, variable) from the [`iris`](https://en.wikipedia.org/wiki/Iris_flower_data_set) dataset which is included by default in R. 
 
 
 ### Fitting the model
 
 
-Let's start by fitting the **frequentist** version of the model, just to have some **reference**:
+Let's start by fitting the **frequentist** version of the model, just to have a reference point:
 
 ```{r message=FALSE, warning=FALSE, eval=TRUE}
 model <- lm(Sepal.Length ~ Petal.Length, data=iris)
 summary(model)
 ```
 
-
-In this model, the linear relationship between **Petal.Length** and **Sepal.Length** is **positive and significant** (beta = 0.41, t(148) = 21.6, p < .001). This means that for every increase of 1 in **Petal.Length** (the predictor), **Sepal.Length** (the response) increases by **0.41**. This can be visualized by plotting, using the `ggplot2` package, the predictor values on the `x` axis and the response values as `y`:
+In this model, the linear relationship between `Petal.Length` and `Sepal.Length` is **positive and significant** (beta = 0.41, *t*(148) = 21.6, *p* < .001). This means that for each one-unit increase in `Petal.Length` (the predictor), you can expect `Sepal.Length` (the response) to increase by **0.41**. This effect can be visualized by plotting the predictor values on the `x` axis and the response values as `y` using the `ggplot2` package:
 
 ```{r message=FALSE, warning=FALSE, eval=TRUE}
 library(ggplot2)  # Load the package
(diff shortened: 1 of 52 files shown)


vignettes/example1.Rmd
```diff
@@ -91,20 +90,21 @@ You can see the sampling algorithm being run.
...
 ```
+
 ```{r message=FALSE, warning=FALSE, echo=FALSE}
 posteriors <- insight::get_parameters(model)
 
 head(posteriors)  # Show the first 6 rows
 ```
 
-As we can see, the parameters take the form of a long dataframe with two columns, corresponding to the **intercept** and the effect of **Petal.Length**. These columns contain the **posterior distributions** of these two parameters, *i.e.*, a set of different plausible values.
+As we can see, the parameters take the form of a wide dataframe with two columns, corresponding to the intercept and the effect of `Petal.Length`. These columns contain the **posterior distributions** of these two parameters. In simple terms, the posterior distribution is a set of different plausible values for each parameter.
```
DominiqueMakowski: actually here the dataframe is "long" (i.e. more rows than columns) ☺️ 

I would also add batsticks around `intercept` as it refers to the parameter name

vignettes/example1.Rmd
```diff
@@ -116,54 +116,57 @@ nrow(posteriors)  # Size (number of rows)
...
 
-
-Now that we've understood where these values come from, let's look at them. We will start by visualizing the distribution of our parameter of interest, the effect of `Petal.Length`.
+Now that we've understood where these values come from, let's look at them. We will start by visualizing the posterior distribution of our parameter of interest, the effect of `Petal.Length`.
 
 
 ```{r message=FALSE, warning=FALSE}
 ggplot(posteriors, aes(x = Petal.Length)) +
-  geom_density(fill = "orange")
+  geom_density(fill = "wheat")
```
DominiqueMakowski: 😄

vignettes/example1.Rmd
```diff
@@ -116,54 +116,57 @@ nrow(posteriors)  # Size (number of rows)
...
 ```
 
 This distribution represents the [probability](https://en.wikipedia.org/wiki/Probability_density_function) (the y axis) of different effects (the x axis). The central values are more probable than the extreme values. As you can see, this distribution ranges from about **0.35 to 0.50**, with the bulk of it being at around **0.41**.
 
 > **Congrats! You've just described your posterior distribution.**
 
-And this is at the heart of Bayesian analysis. We don't need *p*-values, *t*-values or degrees of freedom: **everything is there**, within this **posterior distribution**.
+And this is at the heart of Bayesian analysis. We don't need *p* values, *t* values or degrees of freedom: **everything is there**, within this posterior distribution.
```
DominiqueMakowski: throughout most of the package, we spell *p*-values (I think it is the APA-like spelling), so I suggest keeping the `-`

vignettes/example1.Rmd
```diff
@@ -116,54 +116,57 @@ nrow(posteriors)  # Size (number of rows)
...
 We can now go ahead and **precisely characterize** this posterior distribution.
 
 ### Describing the Posterior
 
-Unfortunately, it is often not practical to report the whole posterior distributions as graphs. We need to find a **concise way to summarize it**. We recommend to describe the posterior distribution with **3 elements**. A **point-estimate**, a one-value summary (similar to the *beta* in frequentist regressions); a **credible interval**, representing the associated uncertainty and some **indices of significance**, giving information about the importance of this effect.
+Unfortunately, it is often not practical to report the whole posterior distributions as graphs. We need to find a **concise way to summarize it**. We recommend to describe the posterior distribution with **3 elements**:
+
+1. A **point-estimate** which is a one-value summary (similar to the *beta* in frequentist regressions).
+2. A **credible interval** representing the associated uncertainty.
+3. And some **indices of significance**, giving information about the relative importance of this effect.
```
DominiqueMakowski: In this context, we can remove the "And" at the beginning of point 3

vignettes/example1.Rmd
```diff
@@ -185,11 +188,11 @@ map_estimate(posteriors$Petal.Length)
 
 ```{r message=FALSE, warning=FALSE}
 ggplot(posteriors, aes(x = Petal.Length)) +
-  geom_density(fill = "orange") +
+  geom_density(fill = "wheat") +
   # The mean in blue
-  geom_vline(xintercept=mean(posteriors$Petal.Length), color="blue", size=1) +
```
DominiqueMakowski: for these vline colors I think it's better to keep raw colours (blue and red) as it is how we refer to them after

vignettes/example1.Rmd
```diff
@@ -257,25 +260,25 @@ model <- stan_glm(weight ~ feed, data=data)
...
 
 > **Is this effect different from 0?**
 
 ### ROPE Percentage
 
 Testing whether this distribution is different from 0 doesn't make sense, as 0 is a single value (*and the probability that any distribution is different from a single value is infinite*). 
 
-However, one way to assess **significance** could be to define an area around 0, which will consider as *equivalent* to zero (*i.e.*, absence of, or negligible, effect). This is called the [**Region of Practical Equivalence (ROPE)**](https://easystats.github.io/bayestestR/articles/region_of_practical_equivalence.html), and is one way of testing the significance of parameters.
+However, one way to assess **significance** could be to define an area around 0, which will consider as *practically equivalent* to zero (*i.e.*, absence of, or negligible, effect). This is called the [**Region of Practical Equivalence (ROPE)**](https://easystats.github.io/bayestestR/articles/region_of_practical_equivalence.html), and is one way of testing the significance of parameters.
```
DominiqueMakowski: I am not sure here, as the practically might be redundant and confusing for non-english (in French, "practically" (pratiquement) refers both to "in a practical manner" and to "almost"). Hence, keeping here only equivalent avoids this problem (and the practically comes afterwards either way.

humanfactors:
Sorry, I'm confused by the presence of 13e920c27827bb60656d50d15b3031c07c37d942 in the PR as that was my fork updating to latest dev.

DominiqueMakowski:
> Sorry, I'm confused by the presence of 13e920c in the PR as that was my fork updating to latest dev.

It's normal, this way there are no conflicts between the branches

vignettes/example1.Rmd
humanfactors: My understanding is that long refers to the format of:

`MEASURE,VALUE` while wide basically means `ID, VAR1, VAR2` with values contained in columns?

I could be wrong, but I don't think it's about length of the total DF? That would be transposition wouldn't it?

vignettes/example1.Rmd
humanfactors: I'm happy to retain the `-` for consistency across the package, though I believe the APA does recommend no hyphen in 6th ed.

vignettes/example1.Rmd
humanfactors: Ahhh I missed this, good spotting.

vignettes/example1.Rmd
DominiqueMakowski: > MEASURE,VALUE while wide basically means ID, VAR1, VAR2 with values contained in columns?

When it refers to repeated measures yes that is correct

But in general a long dataframe can also refer to long in terms of length (number of observations), which is what we wanted to say here (there are 2 cols of 4000 obs)

vignettes/example1.Rmd
DominiqueMakowski: > I believe the APA does recommend no hyphen in 6th ed.

If that is confirmed we will need to slowly remove them in all easystats (as I meticulously placed them everywhere) ^^

vignettes/example1.Rmd
strengejacke: Maybe it's better to call it "large" or something, I found the term "long" also confusing (it somehow implies long versus wide format).

vignettes/example1.Rmd
strengejacke: We could keep practically here, as ROPE also has it in its name. I think here the redundancy serves a better understanding.

vignettes/example1.Rmd
humanfactors: I'll change to lengthy in next commit?

vignettes/example1.Rmd
DominiqueMakowski: (do you have a link where it says that?)

humanfactors: image

I can't actually find an explicit reference in the guide, however, you can see here (APA Style Guide 6th Ed., p. 114) that the guide writes "_p_ value" without the hyphen. However, you do have to have the hyphen for "_t_-test". Nevertheless, I've removed in the commit as it's something we can do a find and replace for at a later date.

vignettes/example1.Rmd
humanfactors: I've kept in f047f6ff655ce87849a2a3ec23e33d88f14342aa

DominiqueMakowski: alright

vignettes/example1.Rmd
DominiqueMakowski: thanks. Yeah, in that case for v3 we will remove the hyphen everywhere ☺️

DominiqueMakowski:
If you're ok @humanfactors I can merge

humanfactors:
All ok and happy to merge!

humanfactors:
Accidental close there....

DominiqueMakowski:
> Feel free to add yourself as ctb in the DESCRIPTION file

Now is the chance to push another commit :)

humanfactors:
Squeezed in the cheeky commit there. Good call!

Issue #168, opened 2019-06-06
p_direction: return as 0.5 - 1 float or keep as 50 - 100 percentage?


strengejacke:
Not sure, probably it's too late, but I just realized that `pd()` returns porportions multiplied by 100, instead, say, .65 for 65%.

Accordingly, `convert_pd_to_p()` requires a number between 1 and 100, where I would expect a number between 0 and 1 (the convention for the probability, I think we had this discussion with ci_level as well...).

Any thoughts? Should we keep the current behaviour, or try breaking changes to the functions?

DominiqueMakowski:
I remember we already had this discussion about pd at the very beginning 😅 , and I think it makes sense here to break this particular convention and force its report as a percentage. Two main reasons being 1) it's less likely to be confused with the other indices (for which the lower the better) and 2) it adds a level of precision (as people usually report two decimals)  which is for pd is particularly relevant.

DominiqueMakowski:
and also it's indeed probably too late :) However, we might maybe emphasize the fact that although it is called probability of direction it is actually usually reported as a percentage. But again the two are the same in the end, and since the print method returns a percentage sign I think it's okay to keep things as they are.

strengejacke:
> and since the print method returns a percentage sign I think it's okay to keep things as they are.

I think, the print-method should do all the job of "transformation", while the raw data is a value between 0 and 1. This will most likely be a confusion if users further process the return-results of `pd()`, or want pipe input into `convert_pd_to_p()`, when these are the only two functions in the world that use "percentage" values instead of their numerical equivalent (proportion) - that is my concern.

DominiqueMakowski:
but `convert_pd_to_p()` works currently appropriately (i.e., it takes in percentages)?

DominiqueMakowski:
I think I would prefer to rename the pd as being the *percentage* of direction than changing the return

strengejacke:
Yes, maybe it's a confusion due to the wording. First, "probability" suggests value between 0 and 1; and second, I can't recall if there are many functions that really take a "percentage" value as input, so typically I would expect and am used to values between 0 to 1 (ratios), rather than "percentages". Even arguments that are labelled as "percentage values" take value between 0 and 1 (see `level` in `confint()`, or `conf.level` in `t.test()`).

I think these points are responsible for my initial confusion here.

DominiqueMakowski:
[...]

PR #194, opened 2019-06-30
dev

DominiqueMakowski:


DominiqueMakowski:
Reading the vignette, I wonder if we should change all mentions of "null *interval*" as "null *region*" to maintain some consistency with ROPE

DominiqueMakowski:
Also null region is less prone to confusion with the uncertainty *intervals* and sounds better than null *interval*... 

(and it makes me think of a hidden valley in the dark forest, right past the misty mountains... the *null* region where unexpected dangers await 🐉)

mattansb:
> (and it makes me think of a hidden valley in the dark forest, right past the misty mountains... the null region where unexpected dangers await 🐉)

I laughed really really hard!

I don't mind null region - it is more consistent with ROPE, and could be clear to users how this is related to ROPE (I took "null interval" from Rouder & Morey).

mattansb:
Actually, changing to region would require a rename of `bayesfactor_interval` to... `bayesfactor_region`?

DominiqueMakowski:
> Actually, changing to region would require a rename of bayesfactor_interval to... bayesfactor_region?

related to the general renaming issue, let me go through to the vignette today maybe I'll have a clearer view and some ideas about an intuitive interface :)

DominiqueMakowski:
If I am correct, the two "main" BF functions are currently `bayesfactor_parameters` and `bayesfactor_models`. Which can both be accessed via the wrapper `bayesfactor()`.

Would it make sense, to avoid the cognitive load of too many different functions, to incorporate bayesfactor_inclusion into `bayesfactor_models` as a bool arg (`inclusion = FALSE`)?

mattansb:
Interesting. That could be done - would simply check if `isTRUE(inclusion)` as pass to `bayesfactor_inclusion`. Is that what you mean?

DominiqueMakowski:
Yup. 

I would then hide the bayesfactor_inclusion, _savagedickey and _interval as internals, favouring their access through the generics

mattansb:
But `bayesfactor_restricted` would still be its own function, correct?

DominiqueMakowski:
Oh right, I forgot about this guy. 

Wouldn't it fit inside bf_parametrers, so that if instead of a scalar or an interval (triggering respectively _SD and _interval), the user passes a string containing the names of parameters, it would run _restricted?

mattansb:
I don't like this - `_restricted` is different from `_parameters` in too many ways:
1. `_parameters` BF is against null / `_restricted` BF is against un-restricted model.
2. `_parameters` BF tests univariate distribution / `_restricted` BF tests multivariate distribution.

It might be in theory closer to `_models` (comparing models), but it really should be used *with* `_models`, and not as a substitute (Maybe I should address this more clearly in the vignette). Anyway, I think this should its own function.

Now that I think of it - I also think `_inclusion` should be its own thing - you would use it for different reasons than `_models`. Additionally, the computation time for `_models` is typically a lot longer than `_inclusion`, and also, you would want to (or you should want to) also *see* the individual models together with the inclusion BFs...

DominiqueMakowski:
I get your point. So that would leave us with 4 different BFs, _models, _parameters, _inclusio & _restriced? The first two possibly triggered via the meta-generic `bayesfactor()`

mattansb:
They can ***all*** be called via `bayesfactor()`!

DominiqueMakowski:
Should we then mainly advertise and showcase the usage of this meta-wrapper? Which makes sense to me, in the end, what is returned are always BFs, indeed they can be very different based on the parameters, hypotheses and models, hence the use of one `bayesfactor()` fun that returns different types based the specified parameters

mattansb:
I'm not a huge fan of the master function - I think it is important for users to be informed about what they're getting, and one way to do that is for the function name to tell them (so they can think - do I want to test parameters? Or the whole model?) - less keen on trusting them to know what they're doing 😅

If I had my way, users should *never* use the master function, and the only reason it even exists is to accommodate `describe_posterior`'s needs. (Like, I would not even export it...)

DominiqueMakowski:
I understand. In that case let's stick to the current structure:

The main functions (i.e., the ones that are likely to be used the most) are bayesfactor_parameters and _models, and then we have additional specific (I'd say *niche*) variants for confirmed users ☺️

mattansb:
Yes - I believe that is why I re structured the vignette (but I don't actually remember haha)

DominiqueMakowski:
> I believe that is why I re structured the vignette (but I don't actually remember haha)

Then maybe we should rearrange the vignette in 3 big sections:

- Testing parameters: if you pass a model to `bayesfactor()`, it tests the parameters. Different options:
  - null region
  - point null
- Comparing models: if you pass multiple models to `bayesfactor()`, it compares them. Different options..
- Alternative use of Bayes factors
  - inclusion: "average" parameter estimation of different models (?)
  - restricted: ~still not clear what it does~ 😅

mattansb:
That's basically the structure now - except `inclusion` and `restricted` are part of the model comparison section... It makes sense to have `inclusion` and `models` together, as `inclusion` will always be preceded by `models`, but I suppose `restricted` can be placed under "even more Bayes factors" or something 😅?

(I still think `bayesfactor()` should not be emphasized to users...) 

What about the `restricted` part was unclear - was the example too vague to follow? (this section was the last part I wrote, and I was getting quite angsty...)

DominiqueMakowski:
> (I still think bayesfactor() should not be emphasized to users...)

👎 😬 

> It makes sense to have inclusion and models together, as inclusion will always be preceded by models

Right, right, makes sense!

> What about the restricted part was unclear 

Must reread it with a cool head :)

DominiqueMakowski:
@mattansb Would it makes sense to rename the `match_models` argument to `match`?

DominiqueMakowski:
Omitting bf_restricted for now:

We have bayesfactor_parameters returns BF related to parameters (as many BFs as they are parameters). bf_models returns BF related to models, as many BFs as models provided (-1).

Then we have bf_inclusion, which takes in a comparison made by bf_models(). To be it would make absolute sense if bf_*parameters* applied to a comparison object did that (too?). 

In other words, could we have:
`bayesfactor_parameters.bayesfactor_models <- bayesfactor_inclusion.bayesfactor_models` 

Then, since we renamed bf_SD to a more generic and abstract name related to its output, I would definitely integrate the current BF_inclusion within the docs of BF_parameters ("if you run this function on a model, you get SD estimates (or BFs interval), but if you run it on a comparison of models (or directly multiple models), you get an inclusion BF). Having these two ways of estimating BFs for parameters together might increase the visibility of BF_inclusion.

mattansb:
> Would it makes sense to rename the match_models argument to match?

Maybe? But why?
(I like that from the name it's clear what is being matched, but this can also be seen in the output I guess...)

---
`parameter` and `inclusion` are too different in their inputs, computation and outputs (e.g., `inclusion`'s BFs are not for parameters, but for terms, they are based on many models, not one...) - would this not confuse users? Even on the most practical level - look at the explanations in the function docs for both - merging them into one would be so convoluted... 
I really need you to explain your philosophy for why less user facing functions is better... I mean, it's not my fault there as so many *different* types of BFs! This would be like merging chi square test, regression analysis, and kolmogorov-smirnov test into the same function just because they all return p-values!

To me, your suggestions just mean that the docs I write are not clear enough on how different these BFs are, and how different their usage is... 😨😢

DominiqueMakowski:
> Maybe? But why?

Well although I am not often pro- short names, here it is kind of obvious that what is matched are the models. The suffix doesn't really clarify or specify the argument (it would have been justified if you *could* match something else, like `match_indices` or whatnot).

> I really need you to explain your philosophy for why less user facing functions is better... 

I must first explain it to myself first 🙄 I don't know really, but I guess that from my personal experience I had a better experience with packages that had just one (or few) main function, that you downloaded to use that particular function. And then by reading the docs of that function, you discovered its power and its potential, that you could keep exploring throughout the years... 

Imagine you read somewhere that the `RING` method is useful for a task x (for instance, the domination of the middle earth). So you download the `mordor` package in which this function is implemented. I would personally be more happy directly typing `mordor::ring()` to see how that works, and then learning about the subtleties, than, when typing `mordor::ring`, having the autocompletion suggesting `mordor::ring_1()`, `mordor::ring_2()`, `...`,  `mordor::ring_9()`, ... Then I have to read the documentation for all variants which ends up with me forfeiting and downloading another package (e.g. the `shire` pkg). 

I *feel* like having fewer but powerful functions helps in terms of visibility rather than mixing everything. It is not a strong opinion though. And that makes me think of the current debate around tidyverse/data.table, in which one of the main criticism of the pedagogical claims of the tidyverse is that it is complicated to learn because there are too many functions, compared to data.table which is super powerful with less names to remember and learn (note that funnily enough, I would say I am currently more on the tidyverse side of things as I think it is a good thing to try to bring programming code closer to natural language, especially in disciplines like psychology)

But it's open for debate, and if you do not feel that we should reduce the number of functions (don't worry I understand, sometimes it can take some time [to let go](https://github.com/easystats/bayestestR/issues/168) 😀), we can still add:

`bayesfactor_parameters.bayesfactor_models <- bayesfactor_inclusion.bayesfactor_models`

and in the docs mention "if you run bayesfactor_parameters on a comparison of models (via bayesfactor_models), the function will compute inclusion BFs" and point toward the inclusion BF docs

(although I understand that inclusion returns terms and not parameters; and I wonder if that would be possible to apply that to actual parameters?)

mattansb:
>Imagine you read somewhere that the RING method is useful for a task x (for instance, the domination of the middle earth). So you download the mordor package...

@DominiqueMakowski I thought you were [a SW fan...](https://www.youtube.com/watch?v=AxAEo3CWeq8)..

I understand what you say about exploring, etc - I think the `bayesfactor()` master function takes care of that - it can be used for everything (if you really really want, we can add at the end of the vignette (in `<sub>` 😅)  a few words about the ~one ring~ master function, we can do that).

{On the tidyverse vs data.table side - I fall on the tidyverse side; though these days I use it less and less (besided `ggplot2`❤), I learned R as a non-programmer (had some basic matlab experience), and data.table seemed like a nightmare (still does, honestly) as I found it easier to google "what function does x?" than "how do I do x with function y?"}

Anyway, I don't feel like we should reduce the number of function (NOR WILL I EVER!). I'll have to think about `bayesfactor_parameters.bayesfactor_models <- bayesfactor_inclusion.bayesfactor_models`...

> I understand that inclusion returns terms and not parameters; and I wonder if that would be possible to apply that to actual parameters?

Unfortunately no - parameters of the same term are fully dependent in which models they are included/excluded, so even if we faked it, all parameters of the same term would have the same BF, which would be very misleading 🤔

strengejacke:
> I feel like having fewer but powerful functions helps in terms of visibility rather than mixing everything.

I would in general agree. However, for this reason, we have chosen to use a "prefix" to "group" function with similar tasks. But I agree that too much diversification can be problematic from a user's perspective. The problem is, there is no definite criterion to decide when to reduce functions, and when to keep them. I haven't followed your discussion in detail because I had some things to get done before vacation... If you haven't desided on this before end of July, I'll jump into this discussion. :-)

DominiqueMakowski:
@strengejacke It might still be going on, it seems that my force is not strong enough... 

image


You might have to use your dark rituals 😅

mattansb:
60691763-96312700-9f04-11e9-84ce-150f7884c1cd

mattansb:
> Would it makes sense to rename the `match_models` argument to `match`?

Should I go a head and do this then? This might break users' code - how would you like this done?

About `bayesfactor_parameters.bayesfactor_models`, I can think of two reason some might accidentally try this:
1. They're trying to test parameters of one of the models in the `bayesfactor_models`.
2. They want an inclusion BFs (for *terms*).

So it might be a good idea to have a function like this, but have it be more of a sign pointing users in the right direction, than having it actually do a thing.

```R
bayesfactor_parameters.bayesfactor_models <- function(...){
  stop(
    "Cannot test parameters of multiple models. You might want one of the following:",
    "\n 1. Passing the model whose parameters you wish to test instead.",
    "\n 2. Using 'bayesfactor_inclusion' to test specific terms across models."
  )
}
```
Or something of the sort... (you'd probably be better at phrasing this...)

DominiqueMakowski:
I am back. We should merge this PR as we agree on the core stuff.

> This might break users' code

This new release will have quite some code-breaking anyway (percentages to proportion and stuff)

> 1. They're trying to test parameters of one of the models in the bayesfactor_models. 

I honestly don't see a user pass by inadvertence multiple models if he's trying to analyze the parameters of one model in particular.

That being said, I am okay with such an informative error :)

I would simplify it to something like `"Oh no, 'bayesfactor_parameters()' does not know how to deal with multiple models :(\nYou want might want to use 'bayesfactor_inclusion()' here to test specific terms across models."`

image

mattansb:
Good to have you back!

I'll get on this + add some missing unit tests for the new functions today, then as far as I'm concerned we can merge.

mattansb:
DONE DONE DONE AND DONE!

@DominiqueMakowski @strengejacke Do you think we should wait with this merge until after the JOSS review has been completed? (Just thinking that some docs would change, etc..)

DominiqueMakowski:
I would rather merge this and open a new PR to address specifically the review. Once everything is accepted, we push this big breaking update to CRAN

DominiqueMakowski:
All looks good to me, but you're sure you don't want to internalize and deprecate the external version of bayesfactor_savagedickey and bayesfactor_inclusion if favour of bayesfactor_parameters? The redundancy is quite confusing IMO 

If not, please do merge whenever you want :)

mattansb:
I kept `bayesfactor_savagedickey` as an alias so as not to break any previous code (and added `bayesfactor_interval` as another alias for the interval method).

Do you think it better to simply have `bayesfactor_savagedickey` give a warning:
```R
warning("bayesfactor_savagedickey is depreciate - use bayesfactor_parameters instead")
```

DominiqueMakowski:
Okay, to deprecate you can directly use the builtin feature in the function: `.Deprecated("bayesfactor_parameters")`

PR #197, opened 2019-07-15
dev

DominiqueMakowski:
This PR will attempt at addressing the reviewer's [comments](https://github.com/openjournals/joss-reviews/issues/1541):

Reviewer 1 (@paul-buerkner)
- [x] **All the citations in the software paper should list a DOI (if they have one) as per reviewer checklist above.**

- Added DOIs for all refs but the following (none was found):
  - `see` package ([here](https://github.com/easystats/see/issues/36))
  - rstanarm package
  - BayesFactor package
  - Mill's "Objective Bayesian Precise Hypothesis Testing" (@tszanalytics do you have a DOI for that?)
  - Multiple Comparisons with BayesFactor, Part 2 (Morey's blog, 2015)
  - Practical bayesian optimization of machine learning algorithms (Snoek's proceedings, 2012)
  - Jeffrey's Theory of Probability book

- [x] **The definition of the maximum a-posteriori value as "the most probable value" is not entirely correct for continuous parameters. Instead it is the value with the highest density (which still have probability zero for continuous parameters).**

- We changed its definition to the following:

"find the **Highest Maximum A Posteriori (MAP)** estimate of a posterior, *i.e.*, the value associated with the highest probability density (the "peak" of the posterior distribution). In other words, it is an estimation of the *mode* for continuous parameters."

- [x] **you seem to use the abbreviation MAP both for the maximum a posteriori and the maximum a-priori value. This will likely confuse readers.**

- This was likely an error and was addressed by replacing instances of the latter by the former (the maximum a posteriori).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,7 @@ Authors@R: c(
 	person("Paul-Christian", 
 		"Bürkner", 
 		role = c("rev"), 
-		email = "paul.buerkner@gmail.com"))
+		email = "paul.buerkner@gmail.com")
 	)
 Maintainer: Dominique Makowski <dom.makowski@gmail.com>
 URL: https://github.com/easystats/bayestestR
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,11 @@ Authors@R: c(
 		"Wilson", 
 		role = c("aut"), 
 		email = "michael.d.wilson@curtin.edu.au", 
-		comment = c(ORCID = "0000-0003-4143-7308"))
+		comment = c(ORCID = "0000-0003-4143-7308")),
+	person("Paul-Christian", 
+		"Bürkner", 
+		role = c("rev"), 
+		email = "paul.buerkner@gmail.com"))
 	)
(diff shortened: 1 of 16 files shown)


mattansb:
@DominiqueMakowski as you've updated the `bayesfactor_savagedickey` to `bayesfactor_parameters` in the paper, should I also update the definition there to the new, broader one?

DominiqueMakowski:
> as you've updated the bayesfactor_savagedickey to bayesfactor_parameters in the paper, should I also update the definition there to the new, broader one?

I think so, no? and in the details section, I would put the differences and specificities of SD and interval (i.e. their respective documentation)

DominiqueMakowski:
> you seem to use the abbreviation MAP both for the maximum a posteriori and the maximum a-priori value. This will likely confuse readers.

@strengejacke @mattansb To address this point, I think the best route is to encompass the "estimate" in the MAP shortcut ("Maximum A Posteriori estimate (MAP)") and to use MAP to refer to the values, instead of the verbose MAP_Estimate (the current column name). As we pretty much always refer to the value throughout the package, it will be more concise and straightforward. Thoughts?

PR #198, opened 2019-07-21
Dev

DominiqueMakowski:


DominiqueMakowski:
```
Dear maintainer,

Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_bayestestR.html>.

Please correct before 2019-08-08 to safely retain your package on CRAN.
```

Related to the recent update of the dataset. Will be fixed once this PR is merged. I am finishing fixing some reordering edgecases in describe_posterior

strengejacke:
Damn, would have been nice to have the updated citation (JOSS) available for the next update.

mattansb:
When are you submitting this?

DominiqueMakowski:
> When are you submitting this?

What do you think? Either we submit this 2.4 now and then will release the v3 after JOSS, or we wait and hope that we will have the reviews in a few days and address them in two weeks (and directly submit the v3).

DominiqueMakowski:
> bayesfactor_restricted <- function(posterior, prior = NULL, hypothesis, verbose = TRUE, ...)

arguments without default should always come first in the args order. Either we should change the order or add a default value.

Also, in BF_parameters, can you remind me why `null` is better than `hypothesis`?

We can merge this PR once it passes travis

mattansb:
> arguments without default should always come first in the args order. Either we should change the order or add a default value.

I've reordered so `hypothesis` is before `prior`.

> Also, in BF_parameters, can you remind me why null is better than hypothesis?

Two reasons:
1. Went back to our original discussion, was never really convinced that the test value was the "hypothesis", but it defninitly does specify the null value(s) (null-point, null-interval).
2. In `bayesfactor_restricted` we *are* specifying hypotheses, and the master `bayesfactor` function needs to distinguish `bayesfactor_restricted` from `bayesfactor_parameters` them 😊

 > What do you think? Either we submit this 2.4 now and then will release the v3 after JOSS, or we wait and hope that we will have the reviews in a few days and address them in two weeks (and directly submit the v3).

I'm fine either way... Hope the JOSS review comes in soon so we won't have to decide 😅

PR #220, opened 2019-09-02
bf_inc error for unsupported mods

strengejacke:
# Description

# Proposed Changes


R/bayesfactor_inclusion.R
```diff
@@ -68,8 +68,8 @@ bayesfactor_inclusion <- function(models, match_models = FALSE, prior_odds = NUL
 
 #' @export
 bayesfactor_inclusion.bayesfactor_models <- function(models, match_models = FALSE, prior_odds = NULL, ...) {
-  if (isFALSE(attr(models, "supported_models"))) {
```
mattansb: I did this for backwards compatibility for object created with previous versions of bayestestR that don't have that attribute. I reverted it back to isFALSE.

PR #243, opened 2019-10-13
dev

DominiqueMakowski:
Introducing new function to convert a Bayesian model into a frequentist one. Currently only support lm/glm and lmer/glmer. Could be improved.

Why? It's part of an idea for https://github.com/easystats/effectsize/issues/8. At first I thought about making it an internal in parameters but then I thought, that this convenience function could be useful for people to run quick comparisons.


DominiqueMakowski:
Can merge then (and submit?)

strengejacke:
I just submitted _insight_ 0.6.0, maybe you want to wait when it's accepted on CRAN, or do you want to submit bayestestR independtly from insight?

DominiqueMakowski:
Can wait so we can update requirements

strengejacke:
last submission wasn't even a month ago, so at some point CRAN will tell you that you should submit no more often than every 1-2 months anyway ;-)

DominiqueMakowski:
this is a major round of update tho, after easystats is fully on CRAN the pace will (*might*) slow down

PR #261, opened 2019-12-02
Dev

DominiqueMakowski:


mattansb:
@DominiqueMakowski why did you add back `hypothesis` to `bf_parameters`? (a deprecated argument) Did removing this break any code? (non that I can see)

DominiqueMakowski:
> Did removing this break any code? (non that I can see)

Every documented argument should be in the usage (i.e., in the function definition), so either we remove it altogether, or we keep it for the next release (and remove it after), but deprecate it as it is now

mattansb:
Hmm.. It's been deprecated for the last few releases - I think we can get rid of it altogether, do you agree?

DominiqueMakowski:
oh right I forgot! in that case yes I'll remove it altogether :)

strengejacke:
Please let us do some brainstorming on the name `mhdior` - @DominiqueMakowski especially regarding you, doesn't it sound like Christian Dior to you?

strengejacke:
all checks pass, ready to merge, I'd say.

mattansb:
Oops, forgot I fixed that bug way-back when. Anyway, the new value is the correct one.

PR #264, opened 2019-12-09
dev

DominiqueMakowski:


DominiqueMakowski:
And on the side, a facetted plot of the evolution of indices of significance:

[](https://www.datanovia.com/en/blog/gganimate-how-to-create-plots-with-beautiful-animation-in-r/)

We could then think about having multiple "evidence accumulation plots" depending on parameters SUCH AS prior specification (we could be cool for papers, as supp material or what)

strengejacke:
?

DominiqueMakowski:
lol I am trying to create a fancy plot that we could use at least in the README and vignettes 😁

mattansb:
That is so cool!!!! We should use that in the priors paper!

DominiqueMakowski:
Thanks @mattansb for a more enthusiastic comment than a depressing "?" 😬

strengejacke:
:scream:
I just successfully reduced package size ~1mb, and now it's being filled with strange animations.

DominiqueMakowski:
okay okay I'll move it to easystats/easystats 😠 😄

DominiqueMakowski:
PS: @mattansb in print.bf_parameters, I swapped the column name as `"Bayes factor"` in favour of `BF` so its consistent with the actual name of the column (just now I got confused by trying to pull a column named "Bayes factor") (and to avoid redundancy as "Bayes Factor" is in the header)

mattansb:
@DominiqueMakowski then I think you might need to also change it in the other `print` method...?

DominiqueMakowski:
which other?

R/print.bayesfactor_parameters.R
```diff
@@ -10,7 +10,7 @@ print.bayesfactor_parameters <- function(x, digits = 2, log = FALSE, ...) {
 
   BFE$BF <- .format_big_small(BFE$BF, digits = digits)
 
-  colnames(BFE)[colnames(BFE) == "BF"] <- "Bayes Factor"
+  colnames(BFE)[colnames(BFE) == "BF"] <- "BF"
```
strengejacke: I think we don't need code line at all in this case?

DominiqueMakowski: actually, yes

mattansb:
@DominiqueMakowski post this gif to twitter!! It's sooooo cooooollll!

DominiqueMakowski:
I am waiting for the final fixed version (it takes ages to compute 😅)

DominiqueMakowski:
Here it is with 60 participants :) 

Can we merge this?

mattansb:
Merge to Twitter? ✔

DominiqueMakowski:
(merged to twitter 😁) Now the PR?

PR #269, opened 2020-01-02
SI methods to `ci()` and `describe_posterior()`

mattansb:
@DominiqueMakowski @strengejacke please take a (quick) look at the implementation (:

### DIFF
--- a/R/si.R
+++ b/R/si.R
@@ -211,7 +211,7 @@ si.data.frame <- function(posterior, prior = NULL, BF = 1, verbose = TRUE, ...){
   out
 }
 
-.si <- function(posterior, prior, BF = 1, extend_scale = 0.05, precision = 2^8) {
+.si <- function(posterior, prior, BF = 1, extend_scale = 0.05, precision = 2^8, ...) {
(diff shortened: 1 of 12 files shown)


R/si.R
```diff
@@ -132,14 +132,16 @@ si.stanreg <- function(posterior, prior = NULL,
                        BF = 1, verbose = TRUE,
```
strengejacke: Looks all good to me! Was just wondering if we should name the argument `BF`, or `bayesfactor`, or both (alias)?

mattansb: Having `bayesfactor` might be more confusing than it already is, as this "BF" is not explicitly ***the*** Bayes factor, but a Bayesian (updating) factor. So I think keeping it as is (`BF`) is best, which also matches the terminology in the SI paper (where the support level is simply rendered as *BF*).

strengejacke: ok, sound feasible!

mattansb: ~So what you're saying is that `F_to_eta2` should be `F_to_eta2_partial`?~
👍

PR #277, opened 2020-01-21
Fix CRAN check issues

strengejacke:


strengejacke:
We should include pkgs like *dplyr*, *ggplot2*, *BFBayesFactor* or *emmeans* etc. as well in the if-statement (depending on which packages are loaded in a vignette), to handle this issue correctly.

strengejacke:
You need to add _scales_ to the suggested packages (and use it conditionally in this vignette) ;-)

mattansb:
removes the use of `scales` instead (:

DominiqueMakowski:
Dear maintainer,

package bayestestR_0.5.1.tar.gz does not pass the incoming checks automatically, please see the following pre-tests:
Windows: <https://win-builder.r-project.org/incoming_pretest/bayestestR_0.5.1_20200125_045322/Windows/00check.log>
Status: 4 ERRORs, 1 WARNING
Debian: <https://win-builder.r-project.org/incoming_pretest/bayestestR_0.5.1_20200125_045322/Debian/00check.log>
Status: 1 ERROR, 1 WARNING, 1 NOTE

--- 
seems related to BF 🤔

mattansb:
Seem like:
- issue with the examples - I wrapped all examples that call other packages in `#' \dontrun{...}`.
- I also made more tests more conditional on other packages not only `rstanarm`.
- Also tried to make the vignette even MORE conditional.
- Issue with `stringi`??? Do we use `stringi`??

I have no idea why the tests failed like they did - they were not suppose to even run :/

mattansb:
> seems related to BF 🤔

Always blaming the BFs.....

---

This whole dependency thing is messed up :/

strengejacke:
> issue with the examples - I wrapped all examples that call other packages in #' \dontrun{...}.

I additionally added a `if (require(...))`, in case we need to run some examples. CRAN sometimes complains if too many examples are wrapped in `dontrun[}`, but for now I preserved all dontruns.

> I also made more tests more conditional on other packages not only rstanarm.
> Also tried to make the vignette even MORE conditional.

I also revised some minor parts to be more safe ;-)

> Issue with stringi??? Do we use stringi??

No, but once other packages are loaded that require stringi, this causes to fail our examples/tests/vignettes as well. I think we now should have fixed everything.

strengejacke:
running local checks now...

strengejacke:
local check OK.

strengejacke:
3n63hz

mattansb:
All of this mess makes for some ugly test/example/vignette code 😔
We might need to think about to incorporate this mess into the easystats style guide?

PR #283, opened 2020-01-31
Dev

strengejacke:


R/bayesfactor_models.R
```diff
@@ -154,28 +154,32 @@ bayesfactor_models.default <- function(..., denominator = 1, verbose = TRUE) {
...
   }
 
+  # Get formula / model names
+  mforms <- mnames
+
   # supported models
   supported_models <- all(sapply(mods, insight::is_model_supported))
   if (supported_models) {
     # Test that all is good:
     resps <- lapply(mods, insight::get_response)
```
strengejacke: You could use `.compact_list(lapply(mods, insight::get_response))` here, and if all results are `NULL` (line 166), you could just check for `length(resps) > 0` and then you are making sure that in line 167, only "valid" elements are sapply'd over.

mattansb: But if even one is missing, I'd want it to give the later warning. Which also means that the `all` should be `any` on line 166. Fixed.

DominiqueMakowski:
```
Dear maintainer,

Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_bayestestR.html>.

Please correct before 2020-02-11 to safely retain your package on CRAN.
```

Although all tests are now marked as OK, should we resubmit nonetheless?

mattansb:
The never ending story...

strengejacke:
> Although all tests are now marked as OK, should we resubmit nonetheless?

No. Else, we have too many submissions, and CRAN also checks for number of submissions per 6 months or so.

strengejacke:
But I would ask if it's now OK to do nothing, as all checks are OK.

DominiqueMakowski:
Okay, will ask.

strengejacke:
Not sure, but I think we have fixed these issue already (https://github.com/easystats/bayestestR/commit/963233673709dee14e596576b50b22847c379702), but not submitted to CRAN yet. I think it should be ok to submit the current master.

strengejacke:
ok, actually, most examples still need to be fixed. even if not run, we could "prepare" them in case some examples should run.

Issue #282, opened 2020-01-28
Brain Bug: logical operations


strengejacke:
@DominiqueMakowski @mattansb Guys, we have made a stupid mistake, meaning that pkgs like bayestestR now still fail ("warn") for CRAN checks...

```if (!requireNamespace("rstanarm", quietly = TRUE) &&
    !requireNamespace("dplyr", quietly = TRUE) &&
    !requireNamespace("ggplot2", quietly = TRUE)) {
  knitr::opts_chunk$set(eval = FALSE)
}
```

**should be**

```
if (!requireNamespace("rstanarm", quietly = TRUE) ||
    !requireNamespace("dplyr", quietly = TRUE) ||
    !requireNamespace("ggplot2", quietly = TRUE)) {
  knitr::opts_chunk$set(eval = FALSE)
}
```

`&&` => `||`

DominiqueMakowski:
What do you mean?

strengejacke:
I think I have fixed all occurrences... In the vignettes, we have `!requireNamespace() && !requireNamespace()` etc. But it should be `||`, not `&&`. Else, the vignettes will be compiled, unless _all_ packages are _unavailable_. However, it should only compile if _all_ packages are _available_.

So, for bayestestR 0.5.1, we now still have warnings on CRAN:
https://cran.r-project.org/web/checks/check_results_bayestestR.html

Which probably should be fixed the next time (we could wait some days for rstanarm to be fixed, then the warnings will disappear).

mattansb:
oops :/

strengejacke:
damn, I hope we have fixed every occurence now, and in the right way... better closing this - at some point, we also need _hope_ to fix bugs. ;-)

PR #284, opened 2020-02-05
Fix CRAN checks

strengejacke:
related to https://github.com/easystats/bayestestR/issues/282

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bayestestR
 Type: Package
 Title: Understand and Describe Bayesian Models and Posterior Distributions
-Version: 0.5.1
+Version: 0.5.2
 Authors@R: c(
     person("Dominique", 
 		"Makowski", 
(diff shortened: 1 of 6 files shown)


DominiqueMakowski:
It passed all CRAN checks, I submitted to CRAN

DominiqueMakowski:
```
package bayestestR_0.5.2.tar.gz does not pass the incoming checks automatically, please see the following pre-tests:
Windows: <https://win-builder.r-project.org/incoming_pretest/bayestestR_0.5.2_20200206_024941/Windows/00check.log>
Status: OK
Debian: <https://win-builder.r-project.org/incoming_pretest/bayestestR_0.5.2_20200206_024941/Debian/00check.log>
Status: 1 WARNING, 1 NOTE
```
Debian fails because of rstanarm

```
* checking package dependencies ... NOTE
Package suggested but not available for checking: ‘rstanarm’
```

What should we do? Report a false positive failure? Resubmit later?

strengejacke:
But that's just the note, not the warning. Where do the check yield the warning?

strengejacke:
Ok, these are the lines that error:

This code chunk (362-367) in _example1.rmd_ fail, because when rstanarm is _not_ available, all chunks are _not_ evaluated. In this case, it's not evaluated and we have no value for `twosided_p`.

```{r message=FALSE, warning=FALSE}
pd <- 97.82
onesided_p <- 1 - pd / 100  
twosided_p <- onesided_p * 2
twosided_p
```

Then, the inline-code at line 369 fails, because it requires `twosided_p`, which is not avaiable

> If we ran our model in the frequentist framework, we should approximately observe an effect with a *p*-value of `r round(twosided_p, digits=3)`. **Is that true?**


Solution: You may try to explicitely set `eval=TRUE` just for the above code chunk. Else... no good idea?

strengejacke:
ok, I think there were still some logical operation issues in the setup-chunk of the vignettes. I also forces the evaluation of the above code chunk.

Issue #285, opened 2020-02-11
download models only for local running tests


strengejacke:
Same as here:
https://github.com/easystats/modelbased/issues/46

strengejacke:
Temporary issue, so closing it

PR #321, opened 2020-06-22
dev

DominiqueMakowski:


mattansb:
@DominiqueMakowski The problem with these test changes is that if I were to fix some bug in `bf_models()` I would have no way of knowing if it broke anything other than the returned data structure. e.g., after a code change, BFs are the log(BF), or BF+1, we would have no way of catching this.

This is why I suggested skipping all the tests on CRAN and only run them locally / continuous integration.

DominiqueMakowski:
Is there an in-between solution? That would allow us to keep the tests while not testing the exact values? but something more sensitive to changes than length)

mattansb:
Why is testing exact values bad if it's only used for development purposes?

This seems like an in-between solution for MCMC-depndant tests: 😅

> .. skipping all the [MCMC-depndant] tests on CRAN and only run them locally / continuous integration.

DominiqueMakowski:
I think testing values in non-fully deterministic process is non recommended because the variability can be induced/increased by unknown forces, in particular hardware and OS differences, and therefore failing tests could be false positives (?). 

Maybe an in-between solution would be like instead of c(a,b,c) == c(x,y,z,), if you know that `b` should be bigger than a and c, have `expect_istrue(all(a > c(b, c)))`, because as the values themselves can change, the "result" (the order) should be less subject to change?

mattansb:
> `expect_istrue(all(a > c(b, c)))`

The problem with this is that this isn't the level of precision we want to guarantee to users. If `ci = 0.8` is more narrow than `ci = 0.9` is that enough? I think we'd also want to see that the returned numbers are reasonable.

> I think testing values in non-fully deterministic process is non recommended because the variability can be induced/increased by unknown forces

True, but if these tests are just for dev use, we wouldn't care about fail/pass, we would carea bout the size of the deviation. If target BF is 3.5 and 6.1, and I get 4.1 and 5.5, that's probably okay, but if I get 11.2 and 41.1 that is probably due to a real error, not just due to any uncertainty.

Having these tests used only internally by us, and skipped on CRAN would give us this info.

mattansb:
Another alternative - most functions can take a data.frame.
So we can have two sets to tests:
1. Does the function behave as expected when given data frames (exact values, not given to chance)
2. Are other non-data-frame objects correctly converted to data frames? (structure of data)

What do you think?

mattansb:
Here's an examples:

``` r
library(rstanarm)
library(bayestestR)
library(testthat)


# function
post <- data.frame(x = distribution_normal(1e4))
prio <- data.frame(x = distribution_normal(1e4, mean = 0.5))

BF0 <- bayesfactor_parameters(post,prio)
#> Loading required namespace: logspline

expect_equal(BF0$BF, 0.88, tol = 0.1)


# structure + function, no exact values
set.seed(1)
fit <- stan_glm(mpg ~ ., data = mtcars, refresh = 0)

set.seed(2)
fit_p <- update(fit, prior_PD = TRUE)

set.seed(2)
BF1 <- bayesfactor_parameters(fit)
#> Computation of Bayes factors: sampling priors, please wait...

BF2 <- bayesfactor_parameters(fit, fit_p)

expect_equal(log(BF1$BF), log(BF2$BF), tol = 0.1)
```

<sup>Created on 2020-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DominiqueMakowski:
You mean like testing all functions first with dataframe inputs and testing the values (i.e., testing the "correctness" of the algorithm per se), and then make sure that other input types (models) are correctly "processed"? 

I guess it makes sense in most cases, but rewriting all the tests seem like a big effort ^^, maybe we can start with the top priority one to see how it goes, like one of the bayesfactor_?

mattansb:
I'll do this for the BF stuff (:

mattansb:
Should have done the trick.

strengejacke:
Note that the past days/weeks GitHub had quite a lot server problems. So using circus downloads for tests should be done only locally (see https://github.com/easystats/bayestestR/issues/285)

DominiqueMakowski:
Do you think we can merge and release to CRAN or did I overlooked something?

mattansb:
> So using circus downloads for tests should be done only locally

Maybe we can add the model objects as (internal) objects to the package for this purpose?

> Do you think we can merge and release to CRAN or did I overlooked something?

Tests are failing on windows because `brms` is missing? and on MacOS can't build vignettes?
Other than that, seems good.

DominiqueMakowski:
> Maybe we can add the model objects as (internal) objects to the package for this purpose?

Since models store the data I'm not sure it's necessary, especially since it adds little to the testing coverage. In fact, stan_glm has become quite fast so we can fit them without much overheard, the problem only persists for more complex models, - and brms ones.

GH-CI fails because:

- `Package suggested but not available: 'brms'` (windows)
- `[error]Error: processing vignette 'indicesEstimationComparison.Rmd' failed with diagnostics:
there is no package called 'broom'` which I'm currently trying to fix by replacing broom with parameters

mattansb:
Yes, I was thinking of `brms` models when suggesting that. Though these have also become somewhat faster, have you noticed??

DominiqueMakowski:
i've read somewhere that it was related to some rstan under the hood improvements. The limitation is for small brms models is the C++ transcription, which seems to give an incompressible overhead

strengejacke:
> Maybe we can add the model objects as (internal) objects to the package for this purpose

This would increase the package size quite a lot, so I would rather not do this. Package size is very large already.

strengejacke:
The checks mostly fail due to server errors. I don't think it's related to specific OSses

> Could not fetch https://github.com/easystats/easystats/raw/master/man/figures/bayestestR/deathsticks.jpg

I think these are still related to the ongoing server problems from GitHub, which might be resolved somewhen in the (near) future.

PR #333, opened 2020-08-04
Fixed bug in effective_sample

sam-crawley:
Previously, if a model had only one parameter (e.g. a null model) effective_sample threw an error. This was because the subset was returning a vector rather than a matrix.


strengejacke:
Thanks! Please also add yourself as contributor to the DESCRIPTION, if you like.

strengejacke:
Just reviewed the PR, and looks good! `subset()` is one solution, adding `, drop = FALSE` into the brackets would do the same (and this is what we have done in other places, but forgot here...)

PR #366, opened 2021-01-25
ROPE: add support for brms multivariate models

mutlusun:
# Description

This PR aims at adding support for brms multivariate models in ROPE calculation. Thanks for your nice package!

# Proposed Changes

I changed the `rope.brmsfit` function to support multivariate models.

# Remarks

- `print.rope`: I decided to print an additional column instead of different headlines (like printing different CIs). If there are multiple responses and different CIs, the output could look a bit messy.
- `rope_range`: Returns now a list with named vectors in the case of a multivariate model. I thought this is more user friendly, if manually specified. This is not a breaking change because multivariate models were not supported yet.
- `.prepare_output`: Please have a look at the changes there. You might want a different approach.

Best

### DIFF
--- a/R/rope.R
+++ b/R/rope.R
@@ -325,31 +325,45 @@ rope.stanfit <- rope.stanreg
 
 #' @rdname rope
 #' @export
-rope.brmsfit <- function(x, range = "default", ci = .89, ci_method = "HDI", effects = c("fixed", "random", "all"), component = c("conditional", "zi", "zero_inflated", "all"), parameters = NULL, verbose = TRUE, ...) {
+rope.brmsfit <- function(
+  x, 
+  range = "default", 
+  ci = .89, 
+  ci_method = "HDI", 
+  effects = c("fixed", "random", "all"), 
+  component = c("conditional", "zi", "zero_inflated", "all"), 
+  parameters = NULL, 
+  verbose = TRUE, 
+  ...
+) {
   effects <- match.arg(effects)
   component <- match.arg(component)
 
-  if (insight::is_multivariate(x)) {
-    stop("Multivariate response models are not yet supported.")
-  }
-
+  # check range argument
   if (all(range == "default")) {
     range <- rope_range(x)
+    #print(range)
   } else if (!all(is.numeric(range)) || length(range) != 2) {
     stop("`range` should be 'default' or a vector of 2 numeric values (e.g., c(-0.1, 0.1)).")
   }
 
-  # check for possible collinearity that might bias ROPE
+  # check for possible collinearity that might bias ROPE and print a warning
   if (verbose) .check_multicollinearity(x, "rope")
 
-  rope_data <- rope(
-    insight::get_parameters(x, effects = effects, component = component, parameters = parameters),
-    range = range,
-    ci = ci,
-    ci_method = ci_method,
-    verbose = verbose,
-    ...
-  )
+  # calc rope range
+  if (insight::is_multivariate(x)) {
+
+  } else {
+    rope_data <- rope(
+      insight::get_parameters(x, effects = effects, component = component, parameters = parameters),
+      range = range,
+      ci = ci,
+      ci_method = ci_method,
+      verbose = verbose,
+      ...
+    )
+  }
+
 
   out <- .prepare_output(rope_data, insight::clean_parameters(x))
 
--- a/R/rope.R
+++ b/R/rope.R
@@ -102,6 +102,11 @@
 #' rope(model)
 #' rope(model, ci = c(.90, .95))
 #'
+#' library(brms)
+#' model <- brms::brm(brms::mvbind(mpg, disp) ~ wt + cyl, data = mtcars)
+#' rope(model)
+#' rope(model, ci = c(.90, .95))
+#'
 #' library(BayesFactor)
 #' bf <- ttestBF(x = rnorm(100, 1, 1))
 #' rope(bf)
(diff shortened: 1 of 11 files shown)


DominiqueMakowski:
@strengejacke @mattansb you guys might have more experience with multivariate models can you take a look at this?

mattansb:
Sorry, have 0 experience here...

DominiqueMakowski:
@IndrajeetPatil ?

IndrajeetPatil:
Sorry, I don't have experience with multivariate models either. 😞

strengejacke:
One thing is the technical issue (which we didn't address due to laziness), and this looks good so far (though I would test for some cases before merging). The other point is defining a proper ROPE in case we have joint distributions. Not sure if this is an issue at all, but maybe we can add this point to the docs?

mutlusun:
Dear all,

Many thanks for considering this PR. I agree with @strengejacke , the only "real" statistical part is the definition of the ROPE in a multivariate case. However, I could not find any literature that is discussing this case. Right now it is handled without taking the joint distribution into account. I can add a point to the docs and print an message to the user about this?

@strengejacke : Which technical issue you are talking about? The failing tests? Actually, I had the feeling that the `.prepare_output` function could be simplified. I refrained from that because I don't know the code too good and what it is doing at other places in this package.

Best

R/rope.R
```diff
@@ -325,33 +337,77 @@ rope.stanfit <- rope.stanreg
...
-  }
-
+  # check range argument
   if (all(range == "default")) {
     range <- rope_range(x)
+  # we expect a list with named vectors (length two) in the multivariate case.
+  # Names state the response variable.
+  } else if (insight::is_multivariate(x) && 
+    (!is.list(range) || length(range) != length(insight::find_response(x) ||
+        names(range) != insight::find_response(x)))) {
```
strengejacke, line 361: Not sure if the parenthesis are set correctly here?

I guess it should be

```r
(!is.list(range) || length(range) != length(insight::find_response(x)) || names(range) != insight::find_response(x))
```

R/rope.R
```diff
@@ -325,33 +337,77 @@ rope.stanfit <- rope.stanreg
...
+  # calc rope
+  if (insight::is_multivariate(x)) {
+    dv <- insight::find_response(x)
+
+    # ROPE range / width differs between response varialbe. Thus ROPE is
+    # calculated for every variable on its own.
+    rope_data <- lapply(
+      dv,
+      function(dv) {
+        ret <- rope(
```
strengejacke, line 379: I think this works, but I would not name the function argument `dv`, when the input object is also named `dv`.

R/rope_range.R
```diff
@@ -63,7 +63,12 @@ rope_range.brmsfit <- function(x, ...) {
   information <- insight::model_info(x)
 
   if (insight::is_multivariate(x)) {
-    mapply(function(i, j) .rope_range(i, j), x, information, response)
+    ret <- mapply(function(i, j) .rope_range(x, i, j), information, response)
+
+    # return matrix as named list
+    # see https://stackoverflow.com/questions/6819804/how-to-convert-a-matrix-to-a-list-of-column-vectors-in-r
+    # this is not the fastest solution bot keeps columns names
+    as.list(as.data.frame(ret))
```
strengejacke, line 71: Can you elaborate why this is necessary?

R/utils.R
```diff
@@ -156,7 +161,7 @@
   temp$.roworder <- 1:nrow(temp)
   out <- merge(x = temp, y = cleaned_parameters, by = merge_by, all.x = TRUE)
   # hope this works for stanmvreg...
-  if (isTRUE(is_stan_mv) && all(is.na(out$Effects)) && all(is.na(out$Component))) {
+  if (isTRUE(is_stan_mv) || isTRUE(is_brms_mv) && all(is.na(out$Effects)) && all(is.na(out$Component))) {
     out$Effects <- cleaned_parameters$Effects[1:nrow(out)]
```
strengejacke, line 165: I think the two ORs conditions need to be surrounded by `()`.

strengejacke (changes requested):

strengejacke:
> Which technical issue you are talking about?

No, I just meant that the ROPE for multivariate response models needed some change / additions in the code-base that requires a bit thinking (what you have done), and the main reason we didn't already have this was laziness. :-)

R/rope.R
mutlusun: Good catch, thank you!

R/rope.R
mutlusun: Thanks, fixed.

R/rope_range.R
mutlusun: Actually, this is not really necessary. However, I decided to do this to have more user friendly arguments.

The above code returns a matrix, a column for every dependent variable and two rows for the lower and upper ROPE limit. In the case that a user specifies own ROPE limits, they would need to create a matrix with these rows and columns. Most of the time, I find this cumbersome and error-prone because I need to create the matrix and pay attention to build the correct columns and rows. I thought it would be easier to use a list with named vectors which correspond to the dependent variables (comparable to `contrasts` argument in a `lm` function).

In addition, I wanted to keep the return value here and the arguments in the `rope` function consistent. So I decided to change the return value here. However, it would be possible to convert a user-specified list to a matrix as well ...

R/utils.R
mutlusun: Again, many thanks!

mutlusun:
Thank you for your feedback! :-)

In a few minutes I will push a commit to fix this points. From my side, it is ready to be merged.

R/rope_range.R
strengejacke: ok, got it. thanks!

strengejacke (approved):

strengejacke:
Let's wait for the check results, maybe we can spot some issues or see if failing checks have other reasons.
@DominiqueMakowski do you want to look at this PR as well?

DominiqueMakowski:
Can we merge that?

strengejacke:
Yes, I think so.

Issue #365, opened 2021-01-23
converting CI columns in output dataframes to a fraction


IndrajeetPatil:
As decided in https://github.com/easystats/effectsize/issues/280#issuecomment-765327512

strengejacke:
@DominiqueMakowski This would have been nice to have in the current update, but I guess it's too late?

strengejacke:
I think we need to fix ROPE_CI as well...

mattansb:
Has this issue been fully addressed?

strengejacke:
Yes, I think so... @IndrajeetPatil @DominiqueMakowski ? Any ideas?

IndrajeetPatil:
> I think we need to fix ROPE_CI as well...

This is fixed.

``` r
library(bayestestR)
#> Note: The default CI width (currently `ci=0.89`) might change in future versions (see https://github.com/easystats/bayestestR/discussions/250). To prevent any issues, please set it explicitly when using bayestestR functions, via the 'ci' argument.
library(rstanarm)
#> Loading required package: Rcpp
#> This is rstanarm version 2.21.1
#> - See https://mc-stan.org/rstanarm/articles/priors for changes to default priors!
#> - Default priors may change, so it's safest to specify priors, even if equivalent to the defaults.
#> - For execution on a local, multicore CPU with excess RAM we recommend calling
#>   options(mc.cores = parallel::detectCores())
model <- stan_glm(mpg ~ wt + gear, data = mtcars, chains = 2, iter = 200, refresh = 0)
#> Warning: Bulk Effective Samples Size (ESS) is too low, indicating posterior means and medians may be unreliable.
#> Running the chains for more iterations may help. See
#> http://mc-stan.org/misc/warnings.html#bulk-ess
#> Warning: Tail Effective Samples Size (ESS) is too low, indicating posterior variances and tail quantiles may be unreliable.
#> Running the chains for more iterations may help. See
#> http://mc-stan.org/misc/warnings.html#tail-ess

as.data.frame(rope(model, ci = c(.90, .95)))
#> Warning: Note that the default rope range for linear models might change in
#> future versions (see https://github.com/easystats/bayestestR/issues/364).Please
#> set it explicitly to preserve current results.
#>     Parameter CI   ROPE_low ROPE_high ROPE_Percentage Effects   Component
#> 1 (Intercept) 90 -0.6026948 0.6026948       0.0000000   fixed conditional
[...]

PR #367, opened 2021-01-26
CI as fraction #365

strengejacke:
See https://github.com/easystats/bayestestR/issues/365

### DIFF
--- a/R/eti.R
+++ b/R/eti.R
@@ -222,7 +222,7 @@ eti.BFBayesFactor <- function(x, ci = .89, verbose = TRUE, ...) {
   ))
 
   data.frame(
-    "CI" = ci * 100,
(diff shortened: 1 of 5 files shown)


strengejacke:
Not sure if all `print()` methods are fully revised now... We also need to convert ROPE_CI, and in general, we could revise all `print()` methods, adding a `format()` method and make use of  `insight::format_table()` and `insight::export_table()` - but this would be another issue. For now, this PR should focus on turning CIs into fractions.

IndrajeetPatil (commented):
All of these changes look good to me!

IndrajeetPatil (approved):

mattansb (approved):
Looks good.
I agree the ROPE /equi-test functions also would need this updated.

Issue #374, opened 2021-02-17
Argument for pd so that the pd is calculated relative to a non-zero number


nahorp:
**Describe the solution you'd like**
Currently, for `pd`, using the `direct` method, the % is a reflection of the ratio of samples superior and inferior to 0 (depending on the median's sign). However, there are occasions, where a non-zero 'reference' value, so to speak, might be beneficial. For instance, passing a contrast that is a ratio (instead of a difference) to `describe_posterior` and hence `pd` (if requested) where 1 indicates 'no difference' and hence that is the value around which the ratio of samples should be calculated (not 0).

Solution would be to be able to choose the reference value (perhaps having 0 as the default)

**How could we do it?**
An argument passed to `pd`, such as, `reference` or `ref`, that can be selected to be 1 (in the above example).

DominiqueMakowski:
Interesting point, @mattansb @strengejacke thoughts? We could indeed add an argument similar to `bf_pointnull(x, null=0)`

mattansb:
Yup, makes sense (:

Are there really any other values that it can take besides 0 (difference) and 1 (for ratio)?

nahorp:
> Are there really any other values that it can take besides 0 (difference) and 1 (for ratio)?

Those are the only 2 that I can think of...

nahorp:
I don't wish to be persistent (considering many aspects that don't require you acting on this request) but is there some sort of timeframe with this or some sort of manual code I could use on the `emmGrid` object that is passed to `describe_posterior` (writing my first Bayesian manuscript based on parameter inference and a bit under the pump to get moving... #phdlife) ?

mattansb:
Until then, here is a quick and dirty work around:

``` r
library(rstanarm)
library(emmeans)

m <- stan_glm(
  am ~ factor(cyl),
  data = mtcars,
  family = binomial(),
  refresh = 0
)

M <- emmeans(m, ~ cyl)
C <- pairs(M)
(Cr <- regrid(C, trans = "response"))
#>  contrast odds.ratio lower.HPD upper.HPD
#>  4/6            3.70     0.146      21.8
#>  4/8           17.76     0.907     120.5
#>  6/8            4.89     0.109      36.3
#> 
#> Point estimate displayed: median 
#> HPD interval probability: 0.95

vals <- insight::get_parameters(Cr)

bayestestR::p_direction(vals - 1)
#> # Probability of Direction (pd)
#> 
#> Parameter |     pd
#> ------------------
[...]

PR #385, opened 2021-03-07
Update p_direction.R

nahorp:
Following on from conversation in issue #374, added a `null` argument for the direct method only (not for density estimation methods) as a proof of concept.

# Description

This PR aims at adding a `null` argument for p_direction as in some instances other values besides 0 are meaningful (e.g., 1 for ratios)

# Proposed Changes

I changed the `p_direction` function so that a `null` argument can be input and only values of 0 (default) or 1 can be input with an error message being printed if values are not 0 or 1.

### DIFF
--- a/R/describe_posterior.R
+++ b/R/describe_posterior.R
@@ -718,7 +718,11 @@ describe_posterior.bamlss <- function(posteriors, centrality = "median", dispers
 
 #' @export
 describe_posterior.get_predicted <- function(posteriors, centrality = "median", dispersion = FALSE, ci = 0.95, ci_method = "hdi", test = NULL, ...) {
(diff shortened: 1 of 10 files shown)


DominiqueMakowski:
@mattansb could you double-check that there is no obvious mistake :)
@IndrajeetPatil is it normal that it doens't pass on the ubuntus and the latest macos?

IndrajeetPatil:
@DominiqueMakowski Yes, the tests are expected to fail on Ubuntu-release and -devel. But, in the PR, the tests are failing for all platforms except two.

Especially, the following failure is not expected and needs to be fixed:

```r
── Error (test-p_direction.R:32:9): p_direction ────────────────────────────────
```

mattansb:
Seems okay to me - just need some tests.

strengejacke:
Why isn't this merged yet?

Issue #420, opened 2021-05-09
For models with very large BFs, bf_inclusion fails to calculate.


Whirly123:
Calculating bf_inclusion fails if models have really large Bayes Factors (eg.  1e+334)

Inclusion BFs will be NA

Using bayestestR_0.8.2  
Models fitted with with BayesFactor package version 0.9.12-4.2

mattansb:
Can you provide a reproducible example?

Whirly123:
> Can you provide a reproducible example?

Here is the reproducible example r code, data attached in a zip because it wasn't accepting my csv for some reason:

```
library(readr)
library(BayesFactor)
library(bayestestR)

data <- read_csv("reproducible.csv")


data$id <- as.factor(data$id)
models <- generalTestBF(Outcome ~ Predictor_A*Predictor_B,
                             whichRandom=c( 
                                      'id', 
                                      'Predictor_A:id',
                                      'Predictor_B:id'
                                           ),
                             data = data)

models

bf_inclusion(models) ### Produces NAs
```

[reproducible.zip](https://github.com/easystats/bayestestR/files/6447367/reproducible.zip)

mattansb:
Yeah, this seems like a precision issue - `10^344=Inf` in R.

You can circumvent it by doing:

```R
models
#> Bayes factor analysis
#> --------------
#> [1] Predictor_A                                         : 6.637297e+227 ±0.01%
#> [2] Predictor_B                                         : 1.117e+336    ±0%
#> [3] Predictor_A + Predictor_B                           : 3.12331e+334  ±0%
#> [4] Predictor_A + Predictor_B + Predictor_A:Predictor_B : 1.175807e+334 ±0.01%
#> 
#> Against denominator:
#>   Intercept only 
#> ---
#> Bayes factor type: BFlinearModel, JZS
 
# Change reference
models <- c(models[-1] / models[1], 1/models[1])
 
models
#> Bayes factor analysis
#> --------------
#> [1] Predictor_B                                         : 1.682914e+108 ±0.01%
#> [2] Predictor_A + Predictor_B                           : 4.705696e+106 ±0.01%
#> [3] Predictor_A + Predictor_B + Predictor_A:Predictor_B : 1.771515e+106 ±0.01%
#> [4] Intercept only                                      : 1.506637e-228 ±0.01%
#> 
#> Against denominator:
#>   Outcome ~ Predictor_A 
#> ---
#> Bayes factor type: BFlinearModel, JZS
 
[...]

PR #421, opened 2021-05-12
Refractor Bayes factors

mattansb:
Return log(BF)

- [x] bf_models
    - [x] as.matrix
    - [x] update
    - [x] format/print
- [x] bf_inclusion
    - [x] format/print
- [x] bf_params
    - [x] format/print
- [x] bf_restricted
    - [x] format/print

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Breaking
 
-- All Bayes factors are now returned as `log(BF)` (column name `log_BF`). Printing is unaffected.
+- All Bayes factors are now returned as `log(BF)` (column name `log_BF`). Printing is unaffected. To retrieve the raw BFs, you can run `exp(result$log_BF)`.
 
 ## Changes
 
@@ -452,4 +452,3 @@
   release](https://github.com/easystats/bayestestR/releases/tag/v0.1.0)
 
 - Added a `NEWS.md` file to track changes to the package
-
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # bayestestR 0.9.0.00001
 
+## Breaking
+
+- All Bayes factors are now returned as `log(BF)` (column name `log_BF`). Printing is unaffected.
+
 ## Changes
 
 * `contr.bayes` has been renamed *`contr.orthonorm`* to be more explicit about its function.
--- a/R/bayesfactor_inclusion.R
+++ b/R/bayesfactor_inclusion.R
@@ -10,7 +10,7 @@
 #' @param prior_odds Optional vector of prior odds for the models. See \code{BayesFactor::priorOdds<-}.
 #' @param ... Arguments passed to or from other methods.
 #'
-#' @return a data frame containing the prior and posterior probabilities, and BF for each effect.
+#' @return a data frame containing the prior and posterior probabilities, and log(BF) for each effect.
 #'
 #' @details Inclusion Bayes factors answer the question: Are the observed data more
 #' probable under models with a particular effect, than they are under models without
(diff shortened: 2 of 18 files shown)


bwiernik:
I think this is a good approach. Printing could show both logBF and BF? Not sure we need a special function to return just the BF considering a user could just do exp(logBF) from the object

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # bayestestR 0.9.0.00001
 
+## Breaking
+
+- All Bayes factors are now returned as `log(BF)` (column name `log_BF`). Printing is unaffected.
```
DominiqueMakowski, line 5:
```suggestion
- All Bayes factors are now returned as `log(BF)` (column name `log_BF`). Printing is unaffected. To retrieve the raw BFs, you can run `exp(result$log_BF)`.
```

copypasta examples ftw 😁

mattansb:
@bwiernik You mean to print both side by side be default?

bwiernik:
I don't know, maybe? What does that look like?

mattansb:
As far as I can tell, everything is working (failed tests have nothing to do with my changes).

@DominiqueMakowski Do you think we should add this info in the start up message?

strengejacke:
How does this affect interpretation rules?

DominiqueMakowski:
> Do you think we should add this info in the start up message?

I would tend to say no, since the change is made explicit in the column name and the printing remains correct

mattansb:
> How does this affect interpretation rules?

Just pass `exp(log_BF)`:

``` r
library(bayestestR)

lm1 <- lm(Sepal.Length ~ 1, data = iris)
lm2 <- lm(Sepal.Length ~ Species, data = iris)
lm3 <- lm(Sepal.Length ~ Species + Petal.Length, data = iris)
lm4 <- lm(Sepal.Length ~ Species * Petal.Length, data = iris)

BFM <- bayesfactor_models(lm1, lm2, lm3, lm4, denominator = lm1)

effectsize::interpret_bf(exp(BFM$log_BF), include_value = T)
#> [1] "no evidence (BF = 1.00) against or in favour of"
#> [2] "extreme evidence (BF = 1.70e+29) in favour of"  
#> [3] "extreme evidence (BF = 5.84e+55) in favour of"  
#> [4] "extreme evidence (BF = 2.20e+54) in favour of"  
#> (Rules: jeffreys1961)
```

<sup>Created on 2021-05-12 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

strengejacke:
Ok, so we _save_ log(BF) in the df, but we _print_ BF, right? Maybe we could add an `as.bayesfactor()` or `as.bf()` method, that just makes `exp(log_BF)`, so users can be sure to have the "correct" BF. I'm not sure why we did this change - just because it fails in certain instances? If we never print log(BF) anyway, we could just return `Inf` and give a warning, saying that some BF are beyond what humans can imagine (and machines can precisely calculate).

DominiqueMakowski:
Also, if it's a problem of memory precision, knowing that the logarithm also tends to infinity there might be such a big BF that even its log is outside the machine precision no? <sub>(it's the magnitude of the BF that 0.95 > 0.89 😬)</sub>

mattansb:
This was does because (1) the lack of precision, and (2) caused issues in `bf_inclusion` even when those resulted in non-astronomical BFs (see #420).

> If we never print log(BF) anyway...

It is an option in the printing functions.

mattansb:
I'm done here. @DominiqueMakowski I think you can merge.

DominiqueMakowski:
does @strengejacke put his veto?

strengejacke:
Go!

PR #426, opened 2021-05-30
to be consistent with other easystats packages

IndrajeetPatil:
Including reviewers in the `DESCRIPTION` file is expected in the context of _rOpenSci_ packages, but not for _JOSS_.

We do not include any of our other _JOSS_ reviewers in other package `DESCRIPTION` files, so we should try to be consistent and fair with our approach across the ecosystem.


IndrajeetPatil:
These same builds are also failing in the `master` branch, so nothing to do with changes I made.

DominiqueMakowski:
@strengejacke @mattansb you validate?

mattansb:
I have no opinion on the matter..

strengejacke:
I personally don't care, but I'm not sure if we should inform paul and tj that we plan to remove reviewers from the description before we actually do it? Doing something "silently" that affects other people is not so kind, even they don't mind?

DominiqueMakowski:
I agree. On top of that, I'm not exactly sure what's the issue with having them here (aside from consistency), is it because it alters the list of contributors / maintainers somehow?

IndrajeetPatil:
Sure.

Or the other option is actually adding our other _JOSS_ reviewers to respective DESCRIPTION files as well.

This way we adopt the same standard across packages. What say you?

DominiqueMakowski:
🤷‍♂️ it's one of the few inconsistencies that legit doesn't bother me at all 😅 
It's just that I feel a bit awkward about removing the credits to someone when it doesn't seem to create a lot of problems

IndrajeetPatil:
> It's just that I feel a bit awkward about removing the credits to someone when it doesn't seem to create a lot of problems

Sure, this is the alternative solution then:
https://github.com/easystats/bayestestR/pull/426#issuecomment-850954774

We should be equal and fair about giving credit to our reviewers.

IndrajeetPatil:
I will make PRs to update DESCRIPTION files for the other packages where I will add JOSS reviewers for those papers.

IndrajeetPatil:
Reviewers added to DESCRIPTION file
- [ ] `effectsize`
- [ ] `correlation`
- [ ] `parameters`
- [ ] `performance`
- [ ] `insight`

PR #450, opened 2021-07-31
model_to_priors()

DominiqueMakowski:
Experimental: nothing to see here ^^

Requires https://github.com/easystats/insight/pull/394


bwiernik:
Ooh. This is cool. It would be great if this could also take metafor objects and produce priors for use in meta-analyses!

DominiqueMakowski:
Have no experience with metafor objects but I don't see why this wouldn't be possible, just need to get the parameters and then re-create a compatible "priors" object, feel free to give it a go if you wish :)

But beware:
image

Some extremist Bayesianists might try to shame us if they see this ^^

Issue #473, opened 2021-10-12
printing for `SI`


mattansb:
@strengejacke You are the printing king.... Can you fix this?

``` r
x1 <- rnorm(1000)
x2 <- rnorm(1000, mean = 0.1, sd = 0.1)

# good
bayestestR::hdi(x1, ci = c(0.89, 0.95))
#> Highest Density Interval
#> 
#> 89% HDI       |       95% HDI
#> -----------------------------
#> [-1.74, 1.38] | [-1.97, 1.86]


# bad
bayestestR::si(x1, x2, BF = c(1 / 3, 1, 3), verbose = FALSE)
#> Support Interval
#> 
#> SI            | CI_low_1 | CI_high_1 | CI_low_3 | CI_high_3
#> -----------------------------------------------------------
#> [-1.74, 1.93] |    -1.74 |      1.93 |    -1.74 |      1.93
```

<sup>Created on 2021-10-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
Ok, the SI only works for one interval. Not sure why, but it must depend on the initial implementation (from bayesfactors?). If I know the relevant use cases, I could fix this in `insight::format_table()`.

mattansb:
*pinging @strengejacke ...*

mattansb:
@strengejacke .....?

mattansb:
This seems to be happening in `insight::format_table()`:

``` r
x1 <- rnorm(1000)
x2 <- rnorm(1000, mean = 0.1, sd = 0.1)

# good
z <- bayestestR::hdi(x1, ci = c(0.89, 0.95))
(z <- datawizard::reshape_ci(z))
#>   CI_low_0.89 CI_high_0.89 CI_low_0.95 CI_high_0.95
#> 1    -1.68565      1.44596   -2.039683     1.911259
insight::format_table(z)
#>          89% CI        95% CI
#> 1 [-1.69, 1.45] [-2.04, 1.91]

# bad
x <- bayestestR::si(x1, x2, BF = c(1 / 3, 1, 3), verbose = FALSE)
(x <- datawizard::reshape_ci(x))
#>   CI_low_0.333333333333333 CI_high_0.333333333333333  CI_low_1 CI_high_1
#> 1                -1.575043                  1.748692 -1.575043  1.748692
#>    CI_low_3 CI_high_3
#> 1 -1.575043  1.748692
insight::format_table(x)
[...]

PR #474, opened 2021-10-12
add warning for inaccurate BF/SI

mattansb:


IndrajeetPatil:
> @IndrajeetPatil Having the paths as they are won't let the vignettes build (see tests).

This is such odd behavior. If you knit the vignette, works without any issues. If you run R CMD Check, it doesn't. You can merge as is and I can have a look at it then. Worse case scenario, we revert to remote URL for CRAN submission.

> Also, why are warnings in the vignette converted to errors???

I think failure to build a vignette is an error, and not a warning? Not sure.

mattansb:
***DO NOT MERGE THIS YET***

mattansb:
Thanks, but this was technically waiting for #473...

IndrajeetPatil:
Oops, _ma culpa_ 🙊

PR #490, opened 2021-11-28
Update bayes_factors.Rmd

jackdolgin:
I addressed the following errors in `bayes_factors.Rmd`
- The prior odds were miscalculated for the sleep example. The prior is said to have sd = 3, but the results are talked about with number 2.5 inserted as the sd (line 222).
- I believe the order of null and alternate probability is flipped in the equations in the sleep example. I believe the alternate probability should be the numerator and the null probability should be the denominator, at least in order to generate the prior odds and posterior odds that get talked about (e.g., "the posterior odds are 2, which seems to favor the effect being non-null"—that statement implies that the alternate probability is in the numerator and the null probability is in the denominator)
- A few typos
- A number of images weren't displaying because their link was wrong

### DIFF
--- a/vignettes/bayes_factors.Rmd
+++ b/vignettes/bayes_factors.Rmd
@@ -39,7 +39,7 @@ knitr::opts_chunk$set(
 
 pkgs <- c(
   "rstanarm", "BayesFactor", "emmeans", "logspline", "lme4", "ggplot2",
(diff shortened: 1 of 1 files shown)


mattansb (approved):

mattansb:
Thanks Jack!

PR #497, opened 2022-01-22
remove unused functions

IndrajeetPatil:


IndrajeetPatil:
The test failures on GHA are the same ones observed in the main/master, so nothing related to this PR.

PR #504, opened 2022-03-02
use insight::get_ll

### LINKED ISSUE #503: Use `insight::get_likelihood` for BIC based BFs

mattansb:
Close #503

@strengejacke will https://github.com/easystats/performance/issues/393 affect `insight::get_loglikelihood`?


strengejacke:
Well, after thinking a bit about it, it will probably indeed affect `get_loglikelihood()`. But is could be handled independently (i,e, `performance_aic()` default REML to FALSE, while `get_loglikelihood()` defaults to the original packages' defaults).

R/bayesfactor_models.R
```diff
@@ -178,7 +176,11 @@ bayesfactor_models.default <- function(..., denominator = 1, verbose = TRUE) {
   }
 
   # Get BF
-  mBIC <- .BIC_list(mods)
+  if (verbose) do.call(BIC, unname(mods))
```
strengejacke, line 179: What's the purpose of this code?

R/bayesfactor_models.R
```diff
@@ -178,7 +176,11 @@ bayesfactor_models.default <- function(..., denominator = 1, verbose = TRUE) {
...
+  mBIC <- sapply(mods, function(m) {
+    LL <- do.call(insight::get_loglikelihood, c(list(m), ll_args))
+    BIC(LL)
```
strengejacke, line 182: `stats::BIC()`

R/bayesfactor_models.R
mattansb: It produces a warning when the models were not fit with the same number of observations

``` r
m1 <- lm(mpg ~ hp, mtcars)
m2 <- lm(mpg ~ hp + gear, mtcars, subset = cyl > 4)

BIC(m1, m2)
#> Warning in BIC.default(m1, m2): models are not all fitted to the same number of observations
#>    df       BIC
#> m1  3 185.63584
#> m2  4  99.89051
```

<sup>Created on 2022-03-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mattansb:
> while `get_loglikelihood()` defaults to the original packages' defaults

Hmmm if this is the case (I'm not sure it is, I'm getting ML LL for `lme4` models), this is not reflected in `get_loglikelihood()`'s documentation.

strengejacke:
> Hmmm if this is the case (I'm not sure it is, I'm getting ML LL for lme4 models), this is not reflected in get_loglikelihood()'s documentation.

No, currently the default is `"ML"`. My thinking was that we _could_ change the default in that way.

R/bayesfactor_models.R
strengejacke: You _could_ use `insight::ellipsis_info()` instead, a function @DominiqueMakowski wrote to evaluate the input of multiple objects.

mattansb: Looking at the internals of that function, it seems to be doing a lot for what I need here...

Perhaps if @DominiqueMakowski would export the `same_response` function?

mattansb:
Waiting to a verdict on get_ll (any changes there will make some tests here fail).

mattansb:
Waiting for tests to fail so I can fix them :P
(When I get back from vacation next week...!)

IndrajeetPatil:
@mattansb Resolved some merge conflicts, so you might want to pull before pushing new changes. Sorry.

mattansb:
Yeah, I'll take care of all of this next week (:

mattansb:
@easystats/maintainers Note that this PR might change results for `bayesfactor_models()` for non-Bayesian models.

This might affect (based on https://github.com/search?q=org%3Aeasystats+bayesfactor_models&type=Code):
- [ ] `performance` @strengejacke 
- [ ] `see` @strengejacke 
- [ ] `report` @DominiqueMakowski

PR #518, opened 2022-04-01
add support for posterior package

strengejacke:
# Description

# Proposed Changes


strengejacke:
@DominiqueMakowski you may merge if everything works.

I mostly copied data frame methods to create the `.draws` methods, so we support the _posterior_ package now.

mattansb (approved):
Looks good to me.

mjskay:
I followed this over from https://github.com/stan-dev/posterior/issues/216. Very cool to see support for {posterior} here!

I noticed that the `.posterior_draws_to_df()` helper doesn't have an implementation for the `"draws_rvars"` format. You should be able to do something like this:

```r
.posterior_draws_to_df.draws_rvars <- function(x) {
  .posterior_draws_to_df(posterior::as_draws_df(x))
}
```

Alternatively you might replace all of the format-specific implementations of `.posterior_draws_to_df()` with something like this:

```r
.posterior_draws_to_df <- function(x) {
  datawizard::data_remove(as.data.frame(posterior::as_draws_df(x)), c(".chain", ".iteration", ".draw"))
}
```

Which would eliminate the need to test/maintain those functions here and make you future-proof to any changes in posterior formats.

strengejacke:
Thanks @mjskay, I used your latter approach, which indeed seems more stable, also regarding potential future changes.

Issue #521, opened 2022-04-04
Default to ETI instead of HDI?


bwiernik:
The more that I'm reading about HDI, the more I'm skeptical about the additional assumptions they require and their performance (especially for >80% intervals). I'm wondering if we should switch to using ETI as the default interval method for Bayesian samples?

> the paper illustrates the challenge of reducing the variance without additional assumptions about the target distribution. After the paper had been published, Andrew started to think that HDI is not that useful, and thus hasn't advocated the paper and the package. In that paper the focus was on 95% HDI, and since then Andrew has advocated, e.g. 20% and 80% quantiles as they have less variance.

__Originally posted by @avehtari in https://github.com/stan-dev/posterior/issues/216#issuecomment-1074177572__

Also: https://twitter.com/betanalpha/status/1479106549519831040?s=21&t=-bBVfRH2pajMzY_ZkqY1Ng

> Highest density intervals are not defined by expectation values and so they cannot be constructed from samples alone.  Like density estimators they require additional assumptions that can significantly affect the outcome of the intervals, especially if the credibility is large.

> Because of this sensitivity to additional assumptions, in particular assumptions that are often only implicit with in a particular HDI estimation method,  I strongly recommend that they be avoided entirely.

strengejacke:
btw, @DominiqueMakowski, here's an argument for 89% instead of 95% interval range ;-)

mattansb:
![image](https://user-images.githubusercontent.com/35330040/161490860-ca50f8d2-4d32-46d8-9714-cb471e22d3d6.png)

bwiernik:
No, those will be about the same. You need to get down to about 60% before the variance is meaningfully smaller.

strengejacke:
That's a strong opinion about HDI:

>  I strongly recommend that they be avoided entirely.

(https://twitter.com/betanalpha/status/1479107186030624771)

strengejacke:
#522 implements SPI (shortest probability intervals)

strengejacke:
Any reason for closing this?

bwiernik:
It was done wasn't it?

strengejacke:
In parameters, not yet bayestestR I think

bwiernik:
Oops

strengejacke:
@DominiqueMakowski any opinions? And: have you submitted _bayestestR_ the last time, and it got suck somewhere? If so, we probably resolve this issue and then submit (again) to CRAN?

[...]

Issue #519, opened 2022-04-04
Documentation of `hdi()` 


strengejacke:
See discussion at https://github.com/stan-dev/posterior/issues/216. There are several algorithms how to compute HDIs, and we're not explicit in our docs which one we use in _bayestestR_.

PR #522, opened 2022-04-04
implement SPI

strengejacke:
#521 

This does not yet change all the docs, I think this would result in too many conflicts, probably. So we might mention the "SPI" option in the docs later.


bwiernik:
Do we want to copy the function from SPIn rather than soft-importing, considering that the package hasn't been updated since 2013?

strengejacke:
I wasn't sure about that, but this can be done easily. We just need the code in `.spi()`. It's more a moral question to me about copying the code 1:1 ;-)

strengejacke:
That method or the code doesn't seem to be very robust:

``` r
library(bayestestR)
x <- rnorm(1000)
spi(x) |> as.data.frame()
#>     CI    CI_low  CI_high
#> 1 0.95 -1.851012 2.079607
SPIn::SPIn(x)$spin
#> [1] -1.851012  2.079607

m <- insight::download_model("brms_1")
d <- as.data.frame(m)
sapply(d, spi)
#> Error in quadprog::solve.QP(D.u, d.u, A.u, c(1, rep(0, u.u - u.l + 2)), : constraints are inconsistent, no solution!

sapply(d, SPIn::SPIn)
#> Error in solve.QP(D.u, d.u, A.u, c(1, rep(0, u.u - u.l + 2)), u.u - u.l): constraints are inconsistent, no solution!
```

<sup>Created on 2022-04-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
The constraints are the `A.l` matrix. I would need to look more into the code for how this is generated and the paper this is based on to understand what's going on

DominiqueMakowski:
should I submit without this PR?

DominiqueMakowski:
> should I submit without this PR or wait for a bit?

*bump* 😬

strengejacke:
w/o this PR, I'd say we must first take a closer look and see how this function can be made more stable. The original `SPIn()` in the _SPIn_ package has some more arguments, which might help adjusting the algorithm to not fail in the above example - but it doesn't help user for whom it is probably still too opaque at the moment what they could do.

strengejacke:
But you _could_ elaborate a bit on https://github.com/easystats/bayestestR/issues/519 :-)I

bwiernik:
@strengejacke The issue that's causing it to fail is the bandwidth argument `bw` is somewhat too large for the intercept and lp_ parameters. 

Let's try this:
1. Use `tryCatch()` with the function using the default `bw = round((sqrt(n.sims) - 1)/2)`
2. If that errors, try again multiplying that bw by .9, .8, then .7

That should make it safe for most distributions.

bwiernik:
(I'll take care of that)

bwiernik:
```r
library(bayestestR)
m <- insight::download_model("brms_1")
d <- as.data.frame(m)
sapply(d, spi)
# > sapply(d, spi)
#         b_Intercept b_wt      b_cyl      sigma    lp__     
# CI      0.95        0.95      0.95       0.95     0.95     
# CI_low  36.25617    -4.703458 -2.386442  1.985231 -83.11161
# CI_high 43.31241    -1.567123 -0.6765242 3.385691 -77.98328
```

bwiernik:
Assuming the checks pass, we can include this

strengejacke:
@DominiqueMakowski have you already submitted? If yes, it's no big deal, but if not, we may include this PR as well - we still need to add this to the docs as well.

strengejacke:
I don't know, this test seems to me like the results are a bit random...

https://github.com/easystats/bayestestR/runs/5856422169?check_suite_focus=true#step:10:336

bwiernik:
That's seems very random. We might want to try to stabilize the seed on that test? But seems unrelated to the current changes

R/spi.R
```diff
@@ -249,9 +253,10 @@ spi.get_predicted <- function(x, ...) {
 
 
   # lower bound
-  if (k == 1) {
+  if (!is.na(k) && all(k == 1)) {
     x1 <- l
     w.l <- 1
+    x.l <- NA
```
strengejacke, line 259: @bwiernik There are some edge cases, where this if-statement is TRUE, and then `x.l` was not initialized. I set it to `NA`, is that ok?

bwiernik:
What if statement do you mean?

strengejacke:
We have

```r
  # upper bound
  if (!is.na(ui) && all(ui == n.sims)) {
    x2 <- u
    w.u <- 1
    x.u <- NA # <- this here is new
  } else {
```

and at the bottom of the function:

```r
  # output
  data.frame(
    "CI" = ci,
    "CI_low" = x.l,
    "CI_high" = x.u
  )
```

when `if (!is.na(ui) && all(ui == n.sims)) {` was TRUE, `x.u` wasn't initialized, so I added `x.u <- NA # <- this here is new` to the if-statement. I also did some minor fixes based on the tests I added, and fixed all errors/issues I found. Mostly rare edge cases, as you can see from the tests. But in general, it now works well with models.

strengejacke:
See my last changes here: https://github.com/easystats/bayestestR/pull/522/commits/a22371d74e8e7544b4cd4ad0bfec7cff94e6af65
Based on the tests I copied from the test-hdi.R file. This showed some issues I fixed in that commit.

strengejacke:
@IndrajeetPatil why does test-coverage always fail? Is there something wrong with the script/action?

bwiernik:
Oh, no. The `x1` and `x2` variables assigned above should just be changed to `x.l` and `x.u`. Missed updating those

strengejacke:
Cool, I think we're set.

bwiernik:
Not sure what's up with those tests

PR #523, opened 2022-04-19
#521

strengejacke:
# Description

# Proposed Changes


strengejacke:
Doing this for `equivalence_test()` is probably a larger breaking change, most likely breaks stuff in _see_.

strengejacke:
We need to fix several tests where the CI output now changes. Simplest way would be to explicitly add `ci_method = "hdi"`, so results won't change.

strengejacke:
close #521

PR #524, opened 2022-04-20
Vignettes online

strengejacke:
# Description

# Proposed Changes


strengejacke:
This PR removes the vignettes and the related figures/pictures from the package tarball, so the file size is reduced by approx. ~90% (down from ~5MB to ~500kb). This PR is built on #523, thus it looks like more changed files than actually are. It's mostly about changing rbuildignore, removing non-used figures and fixing missing bib-entries in readme.

Issue #550, opened 2022-06-26
Replace `{GGally}` with `{bayesplot}`


IndrajeetPatil:
As suggested [here](https://github.com/easystats/easystats/issues/237#issuecomment-1155160641)

IndrajeetPatil:
Admittedly, _bayesplot_ is not lighter than _GGally_:

``` r
tools::package_dependencies("GGally", recursive = TRUE)
#> $GGally
#>  [1] "ggplot2"      "dplyr"        "forcats"      "grDevices"    "grid"        
#>  [6] "gtable"       "lifecycle"    "plyr"         "progress"     "RColorBrewer"
#> [11] "reshape"      "rlang"        "scales"       "tidyr"        "utils"       
#> [16] "generics"     "glue"         "magrittr"     "methods"      "R6"          
#> [21] "tibble"       "tidyselect"   "vctrs"        "pillar"       "ellipsis"    
#> [26] "digest"       "isoband"      "MASS"         "mgcv"         "stats"       
#> [31] "withr"        "Rcpp"         "hms"          "prettyunits"  "crayon"      
#> [36] "farver"       "labeling"     "munsell"      "viridisLite"  "purrr"       
#> [41] "cpp11"        "pkgconfig"    "graphics"     "nlme"         "Matrix"      
#> [46] "splines"      "colorspace"   "cli"          "fansi"        "utf8"        
#> [51] "lattice"

tools::package_dependencies("bayesplot", recursive = TRUE)
#> $bayesplot
#>  [1] "dplyr"          "ggplot2"        "ggridges"       "glue"          
#>  [5] "posterior"      "reshape2"       "rlang"          "stats"         
#>  [9] "tibble"         "tidyselect"     "utils"          "generics"      
#> [13] "lifecycle"      "magrittr"       "methods"        "R6"            
#> [17] "vctrs"          "pillar"         "digest"         "grDevices"     
#> [21] "grid"           "gtable"         "isoband"        "MASS"          
#> [25] "mgcv"           "scales"         "withr"          "plyr"          
#> [29] "abind"          "checkmate"      "tensorA"        "distributional"
#> [33] "parallel"       "matrixStats"    "Rcpp"           "stringr"       
#> [37] "ellipsis"       "fansi"          "pkgconfig"      "purrr"         
#> [41] "backports"      "numDeriv"       "farver"         "graphics"      
#> [45] "nlme"           "Matrix"         "splines"        "cli"           
[...]

PR #552, opened 2022-06-26
Swap bayesplot for GGally

bwiernik:
bayesplot is a dependency of many other Suggested packages, so better to use that, and its pairs plots are visually more appealing

Closes https://github.com/easystats/bayestestR/issues/550


IndrajeetPatil:
Thanks!

Issue #578, opened 2022-12-01
Example broken because data moved


etiennebacher:
Trying to fix https://github.com/easystats/easystats/issues/326, it turns out the data used in the following example is not accessible anymore and I can't find it in another place online.

https://github.com/easystats/bayestestR/blob/1b66d6faabc880adcba9b5b2becbc3354c8c3368/R/bayesfactor_restricted.R#L88-L89

Does someone know if this data can be found elsewhere (and if so can this person put a copy in [easystats/circus](https://github.com/easystats/circus))? Otherwise, should this example be removed?

IndrajeetPatil:
cc @strengejacke, @DominiqueMakowski, @mattansb

strengejacke:
Can't find it after a quick search, so let's remove it?

mattansb:
I'll try and find it, and if I can't I'll replace it with another example (:

mattansb:
Hey @richarddmorey, would you be willing to share the `disgust_example` dataset, as seen here https://www.r-bloggers.com/2015/01/multiple-comparisons-with-bayesfactor-part-1/?

richarddmorey:
Sure - it got broken when my website got hacked. I put it in a gist which will be more stable: https://gist.github.com/richarddmorey/494a3cfa4c9871bc49a7503226efdc9e

mattansb:
Thanks! I'm assuming this is simulated data?

richarddmorey:
Yes, it is simulated. You can do what you like with it, no obligations.

PR #579, opened 2022-12-01
Fix bf res example

mattansb:
Hopefully this is a temp fix of #578 until the original data can be recovered.


etiennebacher:
I guess this can be closed since there's now a link to the old data in #578?

mattansb:
I've restored the original example with Richard's dataset - so as far as I'm concerned this can be merged.

IndrajeetPatil (approved):
Thanks!

PR #596, opened 2023-03-27
Tests cleanup

etiennebacher:
1. Silence messages, cf https://github.com/easystats/easystats/issues/338

2. Use `skip_if_not_or_load_if_installed()` inside `test_that()` calls, cf https://github.com/easystats/easystats/issues/350. There is a test file, `test-emmGrid` where a bunch of `testthat()` all depend on the same model so it feels weird to re-estimate the same model again and again. Also, we can't put all of these in one big `test_that()` because some are skipped on CRAN or else, but not all. So I didn't modify this file.

I have a few failures and warnings locally but I don't know why, let's see on GHA.


IndrajeetPatil (approved):
That was a lot of work. Thanks a lot, @etiennebacher!

I am well-aware of the second point you make in your comment, and I will create an issue in the meta-package repo soon targeting that point.

Issue #609, opened 2023-06-29
Be consistent with return values across functions


strengejacke:
> What ever we decide, we should be consistint - `bayesfactor_parameters()` also returns a single value, but currently returns a data frame. Also, single column data frame inputs to `p_direction()` return a data frame, but a single column matrix returns an scalar.

```R
library(bayestestR)

d <- matrix(rnorm(30))
dim(d)
#> [1] 30  1

p_direction(d) |> str()
#>  'p_direction' num 0.533
#>  - attr(*, "method")= chr "direct"
#>  - attr(*, "data")= num [1:30, 1] -0.913 -0.82 1.518 0.764 -0.415 ...

d <- as.data.frame(d)
dim(d)
#> [1] 30  1

p_direction(d) |> str()
#> Classes ‘p_direction’, ‘see_p_direction’ and 'data.frame':	1 obs. of  2 variables:
#>  $ Parameter: chr "V1"
#>  $ pd       : 'p_direction' num 0.533
#>   ..- attr(*, "method")= chr "direct"
#>   ..- attr(*, "data")= num [1:30] -0.913 -0.82 1.518 0.764 -0.415 ...
```

I think it best to just have tidy-same-format data frame outputs, regardless of input type or size.

__and_ have `as.vector()` or `as.numeric()` methods to "extract" the main index of interest.

_Originally posted by @mattansb in https://github.com/easystats/bayestestR/issues/603#issuecomment-1611299392_

PR #618, opened 2023-07-30
Revise pd

### LINKED ISSUE #616: P-Direction in zero-inflated models
Hello,

i would like to calculate the p-direction of a marginal density that is discontinuous with a large amount of probability mass at exactly zero.

This leads to two problems:
(1) The median is exactly zero, so its unclear whether to use the area to the left or right of zero.
(2) A significant part of the probability mass is exactly zero and neither option from (1) would capture it in calculations.

Currently, the pd() function returns p-direction values below 0.5, violating its theoretical range of 0.5 to 1. 

My question is: Is there any way to solve these two problems without significantly changing the way that the p-direction can be interpreted?

mattansb:
This PR closes #616 and #609.

- [x] Revise p-d docs 
  - [x] How the pd is actually computed
  - [x] possible ranges of values for continuous (0.5-1) vs discretized (0-1) parameter spaces.
  - [x] How values close to 0 can be used to support the null (discretized) hypothesis (https://doi.org/10.1177/2515245921992035)
- [x] Revise pd-to-p 
  - [x] docs: not valid for discretized parameter spaces
  - [x] when pd is smaller than 0.5 we need to return p of exactly 1 with a warning 
  - [x] tests
- [x] For vector inputs:
  - [x] map_estimate returns a data frame
  - [x] p_direction returns a data frame
  - [x] p_map returns a data frame
  - [x] p_significance returns a data frame

### DIFF
--- a/R/p_direction.R
+++ b/R/p_direction.R
@@ -146,7 +146,7 @@ p_direction.default <- function(x, ...) {
 p_direction.numeric <- function(x, method = "direct", null = 0, ...) {
   obj_name <- insight::safe_deparse_symbol(substitute(x))
   out <- p_direction(data.frame(x = x), method = method, null = null, ...)
-  out <- out[-1]
+  out[[1]] <- NULL
   attr(out, "object_name") <- obj_name
   out
 }
--- a/R/p_direction.R
+++ b/R/p_direction.R
@@ -1,81 +1,91 @@
 #' Probability of Direction (pd)
 #'
-#' Compute the **Probability of Direction** (***pd***, also known
-#' as the Maximum Probability of Effect - *MPE*). It varies between `50%`
-#' and `100%` (*i.e.*, `0.5` and `1`) and can be interpreted as
-#' the probability (expressed in percentage) that a parameter (described by its
-#' posterior distribution) is strictly positive or negative (whichever is the
-#' most probable). It is mathematically defined as the proportion of the
-#' posterior distribution that is of the median's sign. Although differently
-#' expressed, this index is fairly similar (*i.e.*, is strongly correlated)
-#' to the frequentist **p-value**.
-#' \cr\cr
-#' Note that in some (rare) cases, especially when used with model averaged
-#' posteriors (see [weighted_posteriors()] or
-#' `brms::posterior_average`), `pd` can be smaller than `0.5`,
-#' reflecting high credibility of `0`.
+#' Compute the **Probability of Direction** (***pd***, also known as the Maximum
+#' Probability of Effect - *MPE*). This can be interpreted as the probability
+#' that a parameter (described by its posterior distribution) is strictly
+#' positive or negative (whichever is the most probable). Although differently
+#' expressed, this index is fairly similar (*i.e.*, is strongly correlated) to
+#' the frequentist **p-value** (see details).
 #'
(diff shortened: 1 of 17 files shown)


mattansb:
The remaining error are unrelated to this PR:
- brms fails to run
- blavaan output is not the right dimension as expected

mattansb:
@DominiqueMakowski I'm done here. Take a look.

R/p_direction.R
```diff
@@ -1,81 +1,91 @@
...
-#' a particular direction (*i.e.*, is positive or negative). Beyond its
-#' simplicity of interpretation, understanding and computation, this index also
-#' presents other interesting properties:
-#' \itemize{
-#'   \item It is independent from the model: It is solely based on the posterior
-#'   distributions and does not require any additional information from the data
-#'   or the model.
-#'   \item It is robust to the scale of both the response variable and the predictors.
-#'   \item It is strongly correlated with the frequentist p-value, and can thus
+#' from 0 to 1, representing the certainty with which an effect goes in a
```
DominiqueMakowski, line 21: isn't it confusing to say that in ranges from 0-1 when in practice it's 0.5-1?

R/p_direction.R
```diff
@@ -1,81 +1,91 @@
...
+#' particular direction (*i.e.*, is positive or negative / has a sign). Beyond
+#' its simplicity of interpretation, understanding and computation, this index
+#' also presents other interesting properties:
```
DominiqueMakowski, line 24: Any reasons to remove:

- It is independent from the model: It is solely based on the posterior distributions and does not require any additional information from the data or the model.

DominiqueMakowski (approved):

R/p_direction.R
mattansb: It is not clear to me what that means - all posterior related functions (so all but BF and CI) are objective descriptions of the posterior - divorced from the data or model.

R/p_direction.R
mattansb: In _your_ practice 😉
I've already come across several applied examples with discrete parameter spaces in my work.

So how about this:
> ... is an index of effect existence, representing the certainty with which an effect goes in a particular direction (*i.e.*, is positive or negative / has a sign), typically ranging from 0.5 to 1 (but see next section for cases where it can range between 0 and 1).  ...

R/p_direction.R
DominiqueMakowski: yes, it's not a unique feature, but it doesn't harm to mention it as well as part of the "interesting properties" (for good or for bad) 🤷

R/p_direction.R
DominiqueMakowski: sounds good, I'll make the edits and merge, thanks a lot!

mattansb:
I think we can merge.

strengejacke:
This PR wasn't ready, and breaks parts of the code, e.g.

``` r
library(bayestestR)
x <- distribution_normal(4000)

describe_posterior(
  x,
  centrality = "all",
  dispersion = TRUE,
  test = "all",
  ci = 0.89,
  verbose = FALSE
)
#> Error in fix.by(by.y, y): 'by' must specify a uniquely valid column
```

<sup>Created on 2023-07-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Because those functions expect numerics instead of data frames for `pd()`, I guess.

PR #622, opened 2023-09-07
Small bug in `contr.equalprior_pairs`

singmann:
A contrast function need to return the contrasts. Currently `contr.equalprior_pairs` does not return anything.


mattansb:
Oops, thanks @singmann !

Issue #619, opened 2023-07-31
bayestestR broken after revising `pd()`.


strengejacke:
This PR wasn't ready, and breaks parts of the code, e.g.

``` r
library(bayestestR)
x <- distribution_normal(4000)

describe_posterior(
  x,
  centrality = "all",
  dispersion = TRUE,
  test = "all",
  ci = 0.89,
  verbose = FALSE
)
#> Error in fix.by(by.y, y): 'by' must specify a uniquely valid column
```

<sup>Created on 2023-07-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Because those functions expect numerics instead of data frames for `pd()`, I guess.

_Originally posted by @strengejacke in https://github.com/easystats/bayestestR/issues/618#issuecomment-1658052172_

DominiqueMakowski:
@mattansb would adding `as.numeric()` be sufficient here?

mattansb:
I have no idea how describe posterior does all the merging. I think pd and p-map already have a as.numeric method.

DominiqueMakowski:
I'm on it but it's a weird error

DominiqueMakowski:
```r
describe_posterior(x, test = "p_map")
```
```
Error in fix.by(by.y, y) : 'by' must specify a uniquely valid column
7.
stop(ngettext(sum(bad), "'by' must specify a uniquely valid column",
"'by' must specify uniquely valid columns"), domain = NA)
6.
fix.by(by.y, y)
5.
merge.data.frame(out, test_pmap, by = merge_by, all = TRUE)
4.
merge(out, test_pmap, by = merge_by, all = TRUE) at describe_posterior.R#474
3.
.describe_posterior(posteriors, centrality = centrality, dispersion = dispersion,
ci = ci, ci_method = ci_method, test = test, rope_range = rope_range,
rope_ci = rope_ci, keep_iterations = keep_iterations, bf_prior = bf_prior,
BF = BF, ...) at describe_posterior.R#540
2.
describe_posterior.double(x, verbose = FALSE, test = "p_map") at describe_posterior.R#109
1.
describe_posterior(x, verbose = FALSE, test = "p_map")
```

it also fails with other tests 🤔 🤔

DominiqueMakowski:
![image](https://github.com/easystats/bayestestR/assets/8875533/743e6413-7778-488a-af5c-b09f40cd2307)

[...]

Issue #603, opened 2023-06-01
Formatting probability of direction objects fails


mjskay:
**Describe the bug**
`format(pd(...))` throws an error, making `pd()` awkward to use in other objects (like data frames or tibbles) which expect `format()` to work.

**To Reproduce**

``` r
set.seed(1234)

x = rnorm(1000)
p = bayestestR::pd(x)
p
#> Probability of Direction: 0.52

format(p)
#> Error in format.p_direction(p): argument "cp" is missing, with no default

df = data.frame(p)
df
#> Error in -digits: invalid argument to unary operator

tbl = tibble::tibble(p)
tbl
#> Error in format.p_direction(x): argument "cp" is missing, with no default

# work around to get p into a data frame: strip attributes
data.frame(p = c(p))
#>       p
#> 1 0.517
```

<sup>Created on 2023-06-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'm not sure if the error thrown by `print(data.frame(p))` is also related to `format()` or something else.

**Expected behaviour**
`format(p)`, `print(data.frame(p))`, and `print(tibble(p))` should all output without error.

**Specifiations (please complete the following information):**
 - tested on most recent CRAN version and most recent Github version as of June 1.

mattansb:
@mjskay note there is a [`as.numeric(<p_direction>)`](https://easystats.github.io/bayestestR/reference/as.numeric.p_direction.html) method.

DominiqueMakowski:
> I think we should change this behavior to be more consistent

I agree yeah

strengejacke:
This behaviour makes (or at least made) sense, because `p_direction()` and the other functions return just one value (for vector inputs), while the other functions at least return two - or, if we have multiple CIs, even more. Thus, the data frame structure was the only way to know which values represents what:

``` r
library(bayestestR)

x <- rnorm(30)
rope(x) |> str()
#> Classes 'rope', 'see_rope' and 'data.frame': 1 obs. of  4 variables:
#>  $ CI             : num 0.95
#>  $ ROPE_low       : num -0.1
#>  $ ROPE_high      : num 0.1
#>  $ ROPE_Percentage: num 0.0714
#>  - attr(*, "HDI_area")='data.frame': 1 obs. of  3 variables:
#>   ..$ CI     : num 0.95
#>   ..$ CI_low : num -1.46
#>   ..$ CI_high: num 2.28
[...]

PR #625, opened 2023-09-29
Rejuvenation

### LINKED ISSUE #525: Inconsistency for name of argument "posterior/s"
See `?estimate_denstity` (`"posterior"`) or `?describe_posterior` (`"posteriors"`). We should use one of the two, not mixing both. Opinions?

@bwiernik @mattansb @DominiqueMakowski

strengejacke:
Fixes #525

Fixing workflows, possibly unresolved issues from #618?

Related to:

- https://github.com/easystats/easystats/issues/318
- https://github.com/easystats/easystats/issues/334
- https://github.com/easystats/easystats/issues/343
- https://github.com/easystats/easystats/issues/348


tests/testthat/test-BFBayesFactor.R
```diff
@@ -19,67 +19,72 @@ test_that("p_direction: BF t.test two samples", {
...
-  expect_equal(as.numeric(p_direction(x)), 91.9, tol = 0.1)
+  expect_equal(as.numeric(p_direction(x)), c(1, 0.95675, 0.95675, 1, 1), tolerance = 0.1)
 })
 
 # BF ANOVA Random ---------------------------
 
 data(puzzles)
 x <- BayesFactor::anovaBF(RT ~ shape * color + ID, data = puzzles, whichRandom = "ID")
 test_that("p_direction", {
-  expect_equal(as.numeric(p_direction(x)), 91.9, tol = 0.1)
```
strengejacke, line 49: @DominiqueMakowski @mattansb @IndrajeetPatil The results of these tests seem to have changed substantially. I hope it is still correct? Not sure whether these were undetected changed in _BayesFactor_ pkg or is there an error we made when addressing #618 / #619 or something?

strengejacke:
@IndrajeetPatil Here are some issues with not available packages, do you know how to fix this?
https://github.com/easystats/bayestestR/actions/runs/6361179508/job/17275434553?pr=625

IndrajeetPatil:
> @IndrajeetPatil Here are some issues with not available packages, do you know how to fix this? https://github.com/easystats/bayestestR/actions/runs/6361179508/job/17275434553?pr=625

Should be fixed for runs from here on.

tests/testthat/test-format.R
```diff
@@ -18,22 +18,22 @@ test_that("p_significance", {
...
   expect_equal(
     format(p_map(x)),
-    data.frame(x = "0.973", stringsAsFactors = FALSE),
+    data.frame(Parameter = "Posterior", p_MAP = "0.973", stringsAsFactors = FALSE),
     ignore_attr = TRUE
   )
   expect_equal(
     format(p_significance(x)),
-    data.frame(x = "0.46", stringsAsFactors = FALSE),
+    data.frame(Parameter = "Posterior", ps = "0.46", stringsAsFactors = FALSE),
```
strengejacke, line 36: @mattansb @IndrajeetPatil @DominiqueMakowski This is how the package behaviour changed due to #603 - please check if this is intentional.

Furthermore, examples have to be revised, see `?p_map`:

`p_map(x, method = "kernel")` -> `as.numeric(p_map(x, method = "kernel"))` in order to work, since numeric inputs now always return a data frame.

tests/testthat/test-BFBayesFactor.R
DominiqueMakowski: I checked and the output of as.numeric is 

```
[1] 1.00000 0.98375 0.98375 0.99375 0.68750 0.83100 0.68325 0.99825 0.66925 0.99900 0.60375 0.99600 0.75925 0.76725 0.99925 1.00000
[17] 1.00000 1.00000
```

I think the test is wrong

DominiqueMakowski: Better to replace with e.g.,:
```r
out <- p_direction(x)
expect_equal(out[out$Parameter == "shape-round", "pd"], 0.98375, tol = 0.1)
```

strengejacke: > I think the test is wrong

Yes, the former test just returned a single value (i.e. one parameter), now we have a bunch of parameters (and values). I changed the test, but what's the reason that the result of the test changed that much? ❓

strengejacke:
Vignette _probability of direction_ is broken. Seems like this code no longer works:

``` r
library(ggplot2)
library(bayestestR)
library(logspline)
library(KernSmooth)
#> KernSmooth 2.23 loaded
#> Copyright M. P. Wand 1997-2009

# Compute the correlations
data <- data.frame()
for (the_mean in runif(25, 0, 4)) {
  for (the_sd in runif(25, 0.5, 4)) {
    x <- rnorm(100, the_mean, abs(the_sd))
    data <- rbind(
      data,
      data.frame(
        direct = as.numeric(pd(x)),
        kernel = as.numeric(pd(x, method = "kernel")),
        logspline = as.numeric(pd(x, method = "logspline")),
        KernSmooth = as.numeric(pd(x, method = "KernSmooth"))
      )
    )
  }
}
data <- as.data.frame(sapply(data, as.numeric))

# Visualize the correlations
bayesplot::mcmc_pairs(data) +
  theme_classic()
#> Warning: Only one chain in 'x'. This plot is more useful with multiple chains.
#> NULL
```

<sup>Created on 2023-09-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/test-BFBayesFactor.R
mattansb: Very weird indeed. Must be a very old test - it also has the probability retuned as a percent value and not a proportion.

strengejacke:
I think we also need updated/new methods for `get_predicted()`, see failing tests here:
https://github.com/easystats/bayestestR/actions/runs/6361855591/job/17276543814?pr=625#step:9:348

tests/testthat/test-format.R
DominiqueMakowski: yes that's correct

strengejacke:
Ok, I think we're coming close to get everything right again. One thing I'm not sure: the `.get_predicted` method, should they return a single value, like

```r
#' @export
p_significance.get_predicted <- function(x, threshold = "default", ...) {
  p_significance(as.numeric(x), threshold = threshold, ...)
}
```

or should they be applied to all iterations, like:

```r
#' @export
spi.get_predicted <- function(x, ...) {
  if ("iterations" %in% names(attributes(x))) {
    out <- spi(as.data.frame(t(attributes(x)$iterations)), ...)
  } else {
    insight::format_error("No iterations present in the output.")
  }
  attr(out, "object_name") <- insight::safe_deparse_symbol(substitute(x))
  out
}
```

strengejacke:
Hm, there were actually quite some bugs in package functions which were not captured by the CI tests? @mattansb Here's an example that seems to have worked in the past (at least, we had a test for it), but now no longer works:

``` r
library(bayestestR)
library(blavaan)
#> Loading required package: Rcpp
#> This is blavaan 0.5-1
#> On multicore systems, we suggest use of future::plan("multicore") or
#>   future::plan("multisession") for faster post-MCMC computations.

data("PoliticalDemocracy", package = "lavaan")

model <- "
  # latent variable definitions
  dem60 =~ y1 + a*y2
  dem65 =~ y5 + a*y6

  # regressions
  dem65 ~ dem60

  # residual correlations
  y1 ~~ y5
"

model2 <- "
  # latent variable definitions
  dem60 =~ y1 + a*y2
  dem65 =~ y5 + a*y6

  # regressions
  dem65 ~ 0*dem60

  # residual correlations
  y1 ~~ 0*y5
"
suppressWarnings(capture.output({
  bfit <- blavaan::bsem(model,
    data = PoliticalDemocracy,
    n.chains = 1, burnin = 50, sample = 100
  )
  bfit2 <- blavaan::bsem(model2,
    data = PoliticalDemocracy,
    n.chains = 1, burnin = 50, sample = 100
  )
}))

bayesfactor_models(bfit, bfit2)
#> Warning: Bayes factors might not be precise.
#>   For precise Bayes factors, sampling at least 40,000 posterior samples is
#>   recommended.
#> Error in apply(diffs, 2, sum): dim(X) must have a positive length
```

<sup>Created on 2023-10-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
Thanks, @strengejacke!

strengejacke:
We should schedule a new release-round... Something like

easystats -> see -> insight -> bayestestR -> ... (parameters? performance?)

strengejacke:
what about _report_, @rempsyc?

datawizard, parameters and performance are quite recent, so probably no update needed shortly.

DominiqueMakowski:
Shouldn't easystats be last to have new versions reqs?

rempsyc:
It's been almost 28 weeks for _report_, I'm happy to resubmit to CRAN. Seems like the devel versions I was relying on were already pushed to CRAN, so technically I could do it next weekend?

strengejacke:
I suggest adding all GitHub version to the REMOTES field, so you can test _report_ on the latest devel versions first - otherwise, we may find revdep issues when we submit packages after _report_ was accepted on CRAN.

rempsyc:
Oh good idea, thanks!

PR #666, opened 2024-08-22
Add `rvar_col` argument to all `*.data.frame()` methods

mattansb:
Fixes #604

```R
grid <- data.frame(
  A = letters[1:2],
  B = c(2, 3),
  val = posterior::rvar(array(rnorm(1200), dim = c(600, 2)))
)

# Pull rvar column:
bayestestR::p_direction(grid$val)
#> Probability of Direction
#> 
#> Parameter |     pd
#> ------------------
#> x[1]      | 53.50%
#> x[2]      | 52.67%

# Or pass the data frame and tell the function what column has rvars:
bayestestR::p_direction(grid, rvar_col = "val")
#>   A B          val        pd
#> 1 a 2 0.062 ± 0.99 0.5350000
#> 2 b 3 0.037 ± 1.00 0.5266667

# Original behavior is maintained when not specifying rvar_col:
bayestestR::p_direction(grid)
#> Probability of Direction
#> 
#> Parameter |   pd
#> ----------------
#> B         | 100%
```

PR #668, opened 2024-08-31
Usability improvements for pd() and pd_to_p()

### LINKED ISSUE #664: Handle missing, NaN, Inf values in pd() and other summary functions
Currently, the posterior summary functions like `pd()` ignore missing and infinite values. These are retained (e.g., as missing) when compared against the `null` value, producing erroneous results like pd = 1 for a vector of all missing values.

```
bayestestR::pd(NA_real_)

x <- c(1, 2, NA, Inf, NaN, -Inf)
bayestestR::pd(x)
```

We should add an `remove_na` argument to these functions, defaulting to TRUE. If `remove_na = FALSE` and there are missing values, the functions should return `NA_real_`. When `remove_na = TRUE`, they should be removed before computing the result. 

For `Inf` and `-Inf` values, arguably these are larger/smaller than 0, but they would usually mean some sort of convergence problem, so I'm thinking it might be better to remove them or to have an argument `remove_inf` defaulting to TRUE to control removing them.

### LINKED ISSUE #665: Usability improvements for pd() and pd_to_p()
I've been using `pd()` a bunch lately, and there are a few things that make it a little annoying to integrate into my workflow. These 2 things would make it much smoother for me:

[ ] A `as_vector` argument that returns a simple vector of pd values, rather than a data frame.

This would fit into a workflow where I have a data frame with posterior draws (e.g., as a list column or as posterior::rvar) and I want to compute pd as a new column. 

(An alternative would be to make the method for `rvar` always return a vector value, but I don't like that inconsistency, and it would mean it's not applicable to similar data structures that don't use `posterior`, like list columns.)

```
results |> transform(pd = pd(.epred))
```

[ ] A `pd_to_p.p_direction()` method that takes the data frame output of `p_direction()` and directly converts it a data frame with p values instead. That would avoid having to do this rather involved set of steps:

```
results |> pd() |> transform(p = pd_to_p(pd)) |> subset(select = - pd)
```

[ ] Maybe it's a little too heretical, but maybe add an `as_frequentist_p` or `as_p` argument to `pd()` to return the results scaled as p values in one step

strengejacke:
Fixes #665
Fixes #664


strengejacke:
@mattansb @bwiernik together with #666 your issue (#665) should be resolved?

strengejacke:
not yet tested, will add some tests.

PR #687, opened 2024-12-16
Multivariate response models

### LINKED ISSUE #31: Multivariate response models
Maybe we manage it to make functions like `equivalence_test()` also work for multivariate response models (like `stanmvreg`), however, this seems more like a long-term goal to me.

strengejacke:
Fixes #31


DominiqueMakowski:
@strengejacke should we merge that before or not?

strengejacke:
Yes, why not? It's only docs that is changed here, but looking at the examples that are added to the docs in this PR, it seems to be working?

PR #695, opened 2025-01-27
`describe_posterior()` only conditionally calls `rope_range()` when needed

mattansb:
Partially address https://github.com/easystats/insight/issues/997


strengejacke:
Looks good! As mentioned in https://github.com/easystats/insight/issues/997#issuecomment-2618085070, I'd also add `no_recursion = TRUE` to all `get_sigma()` calls.

PR #699, opened 2025-02-13
new `rope(x, complement = TRUE)`

mattansb:
New feature to report posterior probabilities above/below the rope for a more full picture.

```R
library(bayestestR)

x <- rnorm(4000, sd = 5)

rope(x, complement = TRUE)
#> # Proportion of samples inside the ROPE [-0.10, 0.10]:
#> 
#> Inside ROPE | Above ROPE | Below ROPE
#> -------------------------------------
#> 1.61 %      |    49.08 % |    49.32 %
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,12 @@
+# bayestestR 0.15.3
+
+## New functionality
+
+* `rope()` (and by extention `p_rope()`) gain a new `complement` argument such
+  that `rope(x, complement = TRUE)` returns the ROPE posterior probability
+  together with the posterior probabilities above/below the ROPE (the 
+  _complementary_ probabilities).
+
 # bayestestR 0.15.2
 
 ## Changes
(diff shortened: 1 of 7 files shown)


DominiqueMakowski:
I'd make that default rather than a toggleable option

mattansb:
Yeah? I thought of this as supplementary...

DominiqueMakowski:
I think it's pretty useful, and it's not a breaking change of the return format since we already return a dataframe so 🤷

If we make it optional it will likely end up in the abyss of forgotten features ^^

mattansb:
Fair enough. Then should this even be optional? Should `complement = FALSE` be an option at all?

DominiqueMakowski:
I'd say no :p 
Let's just document in the docstrings what are these columns and how to interpret them / how useful they can be

strengejacke:
Agree to include it in general and have no option.

strengejacke:
ok, for now, we make it an option. We have to fix `insight::format_table()` to correctly format the new columns

strengejacke:
Mostly, our packages can basically call `insight::export_table(format(x))`, and `format()` does some formatting involving `insight::format_table()`. However, for `rope()`, there's support for multivariate response models, and I haven't checked if it's easily doable with the insight-functions.

PR #705, opened 2025-03-15
effective_sample(): returns tail ESS for brms models

### LINKED ISSUE #704: Describe_posterior(): make it possible to return Bulk and tail ess
I'm on my phone so I can't make an MWE, but essentially brms returns tail and bulk ESS, which can be informative.

In describe_posterior(), setting diagnostic="all" doesn't return them and returns by default another value of ESS

we should provide an option to include the same output as in the default brms printing

DominiqueMakowski:
Fixes #704


DominiqueMakowski:
Didn't find another way to pull ESS than using `posterior`

strengejacke:
I think we also need parameters filtering

strengejacke:
> I think we also need parameters filtering

Oh, it's already there I saw.

strengejacke:
There are a lot of RD-files changed, but that's no problem right now. You said you'd prefer "simple" argument options, so I just kept the main option. We can now leave all the checks for the argument to the _insight_ functions, and furthermore, we thereby also allow all options that are supported in _insight_ (which might have been more there than here in _bayestestR_).

Furthermore, _posterior_ also works for stanreg, so the functions simply differ in their default options, but then the stanreg method just calls the brms method.

We can then later revise all the other bayestestR methods, and only have one option for the `effects` and the `component` argument, and again do all the checks for valid options in _insight_, this simplifies the code and the docs.

PR #706, opened 2025-03-16
Docs, no multiple options in arguments

strengejacke:
This PR removes multiple options for the  `effects` and `component` arguments, leaving only the main option, to make docs clearer.

Furthermore, options are now better documented, and checks are now done in *insight*.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bayestestR
 Title: Understand and Describe Bayesian Models and Posterior Distributions
-Version: 0.15.2.4
+Version: 0.15.2.5
 Authors@R:
     c(person(given = "Dominique",
              family = "Makowski",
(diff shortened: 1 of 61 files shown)


bwiernik:
I'm not following this? The standard approach for R arguments that have a finite list of allowed options is to enumerate them in the function definition like they currently are.

strengejacke:
Yes, the idea was to have only the default option in the docs for the function definition. See discussion here: https://github.com/easystats/modelbased/issues/420#issuecomment-2715800894

strengejacke:
One reason is also because we use `effects` and `component` mostly for calls to `insight::get_parameters()`, where those arguments are evaluated/matched. However, we allow more options in _insight_ now, and instead of updating all functions/methods here, we simply leave this task to _insight_. That's why we just need the "main option" here. And we have a section in the docs that mentions all possibilities.

PR #715, opened 2025-04-23
Make `point_estimate()` robust to constant data

### LINKED ISSUE #714: Make `point_estimate()` robust to constant data
This is a real pain when more higher-level functions (like `describe_distribution()`) calls this function and the user has no recourse to figure out what went wrong, for which grouping level, etc. When the groups are small enough (there being just one observation in a group is a common occurrence), we should anticipate this issue and make sure the function can handle it. 

### Minimal reprex

User view of the error while using a higher-level function:

``` r
library(dplyr)

dplyr::filter(ggplot2::mpg, cyl == "4") %>%
  dplyr::group_by(manufacturer) %>%
  datawizard::describe_distribution(centrality = "MAP")
#> Error in bw_pair_cnts(x, nb, n > nb/2): data are constant in bandwidth calculation
```

<sup>Created on 2025-04-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

After a lot of digging, figuring out which function produces this and for which group:

``` r
library(dplyr)

data <- dplyr::filter(ggplot2::mpg, cyl == "4", manufacturer == "chevrolet")
bayestestR::point_estimate(data, centrality = "MAP")
#> Error in bw_pair_cnts(x, nb, n > nb/2): data are constant in bandwidth calculation
```

<sup>Created on 2025-04-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This can be especially surprising because it does work with other centrality measure and fails only for MAP:

``` r
library(dplyr)

data <- dplyr::filter(ggplot2::mpg, cyl == "4", manufacturer == "chevrolet")
bayestestR::point_estimate(data, centrality = "mean")
#> Point Estimate
#> 
#> Parameter |    Mean
#> -------------------
#> displ     |    2.40
#> year      | 2003.50
#> cyl       |    4.00
#> cty       |   20.50
#> hwy       |   28.50

data <- dplyr::filter(ggplot2::mpg, cyl == "4", manufacturer == "chevrolet")
bayestestR::point_estimate(data, centrality = "median")
#> Point Estimate
#> 
#> Parameter |  Median
#> -------------------
#> displ     |    2.40
#> year      | 2003.50
#> cyl       |    4.00
#> cty       |   20.50
#> hwy       |   28.50
```

<sup>Created on 2025-04-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Fixes #714

### DIFF
--- a/R/map_estimate.R
+++ b/R/map_estimate.R
@@ -57,10 +57,19 @@ map_estimate.numeric <- function(x, precision = 2^10, method = "kernel", ...) {
 }
 
 .map_estimate <- function(x, precision = 2^10, method = "kernel", ...) {
-  d <- estimate_density(x, precision = precision, method = method, ...)
-
-  out <- d$x[which.max(d$y)]
-  attr(out, "MAP_density") <- max(d$y)
+  # sanity check - if we have only one unique value (a vector of constant values)
+  # density estimation doesn't work
+  if (insight::n_unique(x) == 1) {
+    out <- stats::na.omit(x)[1]
+    attr(out, "MAP_density") <- 1
+  } else {
+    d <- try(estimate_density(x, precision = precision, method = method, ...), silent = TRUE)
+    if (inherits(d, "try-error")) {
+      return(NA)
+    }
+    out <- d$x[which.max(d$y)]
+    attr(out, "MAP_density") <- max(d$y)
+  }
   out
 }
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: bayestestR
 Title: Understand and Describe Bayesian Models and Posterior Distributions
-Version: 0.15.2.12
+Version: 0.15.2.13
 Authors@R:
     c(person(given = "Dominique",
              family = "Makowski",
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,9 @@
 * `describe_posterior()` now returns a columns with response variables for
   *marginaleffects* objects applied to multivariate response Stan models.
 
+* Fixed issue in `map_estimate()` and `point_estimate(centrality = "MAP")` for
+  vectors with only one unique value.
+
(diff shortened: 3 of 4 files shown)


R/map_estimate.R
```diff
@@ -57,10 +57,27 @@ map_estimate.numeric <- function(x, precision = 2^10, method = "kernel", ...) {
...
+  # density estimation doesn't work
+  if (insight::n_unique(x) == 1) {
+    out <- stats::na.omit(x)[1]
+    attr(out, "MAP_density") <- 1
+  } else {
+    d <- try(estimate_density(x, precision = precision, method = method, ...), silent = TRUE)
+    if (inherits(d, "try-error")) {
+      if (isTRUE(dots$verbose)) {
+        msg <- "Could not calculate MAP estimate."
+        if (grepl("too sparse", d, fixed = TRUE)) {
```
mattansb, line 71: Data is _maybe_ too sparse gets `NA` + warning,
but singular data (i.e., data that _is_ too sparse) gets singular value and no warning. This feels inconsistent. I this we should add a similar warning (that can be turned off by `varbose=`) for the singular case: "The provided data is a constant. Returning constant."

R/map_estimate.R
```diff
@@ -57,10 +57,27 @@ map_estimate.numeric <- function(x, precision = 2^10, method = "kernel", ...) {
 }
 
 .map_estimate <- function(x, precision = 2^10, method = "kernel", ...) {
-  d <- estimate_density(x, precision = precision, method = method, ...)
-
-  out <- d$x[which.max(d$y)]
-  attr(out, "MAP_density") <- max(d$y)
+  dots <- list(...)
```
mattansb, line 60: Maybe formally add the `verbose=` argument to the function?

mattansb (changes requested):

mattansb:
LGTM!

mattansb (approved):

PR #748, opened 2025-12-02
Fix/rope vignette typo

ethanknights:
# Description

Noticed some minor typos in the ROPE vignette.


strengejacke:
Good catch, thanks!

PR #750, opened 2026-01-20
add basic functions for diagnostic_posterior for lists of chains or 3…

bbolker:
# Description

This PR aims to make the default method of `diagnostic_posterior` work with 'raw' MCMC samples (i.e., lists of data frames or matrices representing samples of parameters from chains, or 3D arrays) as well as objects from `rstanarm`/`brms`/`lavaan` models. This uses `rstan::monitor()`.  As far as I can tell this is theoretically OK, and extends the use of `diagnostic_posterior` to MCMC samples from a wide range of tools.

# Proposed Changes

Changed `diagnostic_posterior.default` to (1) check if input is valid [list of 2D objects with appropriate/matching dimensions, or 3D array]; (2) if it's a list, coerce it to a 3D array; (3) run `rstan::monitor` to get Rhat, bulk ESS, tail ESS (and some other stuff).

Needed improvements:

* probably want to throw away some of the columns returned from `rstan::monitor` (keep only diagnostics)
* this is probably not a good idiom-match to the rest of the code base (but maybe an LLM can handle that??)


strengejacke:
Thanks, looks great! Just two minor comments: 
- don't use tibbles or the tibbles package, use real data frame 
- instead of `stop()` etc. we use `insight::format_error()`

bbolker:
Ironically, the `abind` collision problem bit me in my own code just now. Significantly modified now. Further possible room for improvement: (1) document that MCSE is the MCSE of the median (from MCSE_50 in `monitor` output)? (2) allow possibility of returning Bulk and Tail ESS separately? A wider range of tests (e.g. exercising the input-checking code)?

FWIW I think there's significant scope for refactoring/DRY in the diagnostic-processing code for the other methods, but I decided not to touch that.

Could also, possibly, allow returning ESS and MCSE (only) for input with a single chain (where Rhat can't be computed). Might have to call `coda` instead ... which might not have all the improvements to the underlying definitions that `monitor` uses ...

PR #769, opened 2026-07-03
Fixed .check_multicollinearity util function

will02-code:
# Description

The .check_multicollinearity function previously was setup so that it could never flag probable collinearity (>0.9), becuase the filtering by the first threshold changed the results df. 

# Proposed Changes

Changed so that either threshold can be triggered (created a different df for probable and possible collinearity)

# Notes
First PR so please let me know if I have done anything incorrectly!


strengejacke:
Thanks, looks good to me! Let me also trigger an AI review.

